PR #11, opened 2019-07-20
Correcting repeating prefix length for GoDaddy key

mikefero:
- Operating Test Environment (OTE) key prefixes are 14 characters
- Production key prefixes are 10 characters


qdm12:
Thanks for your correction!

PR #12, opened 2019-08-06
add new dns provider

cnjack:
add dnspod provider(https://www.dnspod.cn/)


.gitignore
```diff
@@ -1,2 +1,3 @@
 *.exe
-.vscode
\ No newline at end of file
+.vscode
+pkg/update/update_test.go
```
qdm12: Why git ignore `pkg/update/update_test.go`?

go.mod
```diff
@@ -12,4 +11,5 @@ require (
 	github.com/mattn/go-isatty v0.0.7 // indirect
 	github.com/mattn/go-sqlite3 v1.10.0
 	github.com/spf13/viper v1.3.2
+	github.com/stretchr/testify v1.2.2
```
qdm12: Do you need testify here?

qdm12:
Thanks for the PR! If you can just review the questions, I'll be happy to merge it.

cnjack:
I use the module in test file. but my password is in the test file, I will move the password to env and push the code tomorrow





------------------ Original ------------------

cnjack:
is there any question?

pkg/update/update.go
```diff
@@ -459,3 +471,93 @@ func updateNoIP(httpClient *http.Client, hostname, username, password, ip string
...
+		if record.Type == "A" && record.Name == host {
+			recordID = record.ID
+			line = record.Line
+			if ip == record.Value {
+				return "", nil
+			}
+		}
+	}
+	if recordID == "" {
+		return "", errors.New("record not found")
```
qdm12: Please use `fmt.Errorf("record not found")` to avoid using the `errors` package on top of `fmt` (more flexible and extensible)

pkg/update/update_test.go
```diff
@@ -0,0 +1,29 @@
+package update
+
```
qdm12: Hi there! 

the test file relies on actual credentials and network connectivity which I would like to avoid for portability and speed of testing. It's a bad coding practice except for database testing.

Please either remove the test file and testify OR rework the test file to use a fake local API server to imitate your DNS provider that would accept any credentials and return the same deterministic response. I think testify has such unit testing ready http server built-in.

pkg/update/update.go
```diff
@@ -459,3 +471,93 @@ func updateNoIP(httpClient *http.Client, hostname, username, password, ip string
...
+	postValues.Set("domain", domain)
+	postValues.Set("length", "200")
+	postValues.Set("sub_domain", host)
+	postValues.Set("record_type", "A")
+	req, err := http.NewRequest(http.MethodPost, "https://dnsapi.cn/Record.List", bytes.NewBufferString(postValues.Encode()))
+	if err != nil {
+		return "", err
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	resp, err := httpClient.Do(req)
```
qdm12: Please use `status, content, err := network.DoHTTPRequest(httpClient, req)` instead

pkg/update/update.go
```diff
@@ -459,3 +471,93 @@ func updateNoIP(httpClient *http.Client, hostname, username, password, ip string
...
+	if err != nil {
+		return "", err
+	}
+	respBody, err := ioutil.ReadAll(resp.Body)
+	resp.Body.Close()
+	if err != nil {
+		return "", err
+	}
+	type RecordResp struct {
```
qdm12: Please use anonymous structs for a simpler code, similarly to [these 5-10 lines](https://github.com/qdm12/ddns-updater/blob/master/pkg/update/update.go#L266)

pkg/update/update.go
```diff
@@ -459,3 +471,93 @@ func updateNoIP(httpClient *http.Client, hostname, username, password, ip string
...
+	resp, err = httpClient.Do(req)
+	if err != nil {
+		return "", err
+	}
+	respBody, err = ioutil.ReadAll(resp.Body)
+	resp.Body.Close()
+	if err != nil {
+		return "", err
+	}
+	type DDNSResp struct {
```
qdm12: use anonynous strucrs here as well, thanks!

pkg/update/update.go
```diff
@@ -459,3 +471,93 @@ func updateNoIP(httpClient *http.Client, hostname, username, password, ip string
...
+			Value string `json:"value"`
+			Name  string `json:"name"`
+		} `json:"record"`
+	}
+	ddnsResp := new(DDNSResp)
+	err = json.Unmarshal(respBody, ddnsResp)
+	if err != nil {
+		return "", err
+	}
+	if ddnsResp.Record.Value != ip {
```
qdm12: Can your DNS provider determine your public IP automatically? Or do have to use another method to obtain your public IP (what seems to be the case from the code). If so, don't return the newIp and only check that the newIp is set to the wanted Ip (already done).

cnjack:
ok, i will fix it tomorrow

----------

该邮件从移动设备发送


--------------原始邮件--------------
发件人："Quentin McGaw "<notifications@github.com>;
发送时间：2019年8月8日(星期四) 晚上7:42
收件人："qdm12/ddns-updater" <ddns-updater@noreply.github.com>;
抄送："王文慧 "<admin@nightc.com>;"Author "<author@noreply.github.com>;
主题：Re: [qdm12/ddns-updater] add new dns provider (#12)
-----------------------------------

 
@qdm12 requested changes on this pull request.
 
 
In pkg/update/update.go:
 > +	if err != nil { +		return "", err +	} +	var recordID string +	var line string +	for _, record := range recordResp.Records { +		if record.Type == "A" && record.Name == host { +			recordID = record.ID +			line = record.Line +			if ip == record.Value { +				return "", nil +			} +		} +	} +	if recordID == "" { +		return "", errors.New("record not found")  
Please use fmt.Errorf("record not found") to avoid using the errors package on top of fmt (more flexible and extensible)
 
 
In pkg/update/update_test.go:
 > @@ -0,0 +1,29 @@ +package update +  
Hi there!
 
the test file relies on actual credentials and network connectivity which I would like to avoid for portability and speed of testing. It's a bad coding practice except for database testing.
 
Please either remove the test file and testify OR rework the test file to use a fake local API server to imitate your DNS provider that would accept any credentials and return the same deterministic response. I think testify has such unit testing ready http server built-in.
 
 
In pkg/update/update.go:
 > @@ -459,3 +471,93 @@ func updateNoIP(httpClient *http.Client, hostname, username, password, ip string  	}  	return "", fmt.Errorf("unknown response: %s", s)  } + +func updateDnsPod(httpClient *http.Client, domain, host, token, ip string) (newIP string, err error) { +	postValues := url.Values{} +	postValues.Set("login_token", token) +	postValues.Set("format", "json") +	postValues.Set("domain", domain) +	postValues.Set("length", "200") +	postValues.Set("sub_domain", host) +	postValues.Set("record_type", "A") +	req, err := http.NewRequest(http.MethodPost, "https://dnsapi.cn/Record.List", bytes.NewBufferString(postValues.Encode())) +	if err != nil { +		return "", err +	} +	req.Header.Set("Content-Type", "application/x-www-form-urlencoded") +	resp, err := httpClient.Do(req)  
Please use status, content, err := network.DoHTTPRequest(httpClient, req) instead
 
 
In pkg/update/update.go:
 > +	postValues.Set("record_type", "A") +	req, err := http.NewRequest(http.MethodPost, "https://dnsapi.cn/Record.List", bytes.NewBufferString(postValues.Encode())) +	if err != nil { +		return "", err +	} +	req.Header.Set("Content-Type", "application/x-www-form-urlencoded") +	resp, err := httpClient.Do(req) +	if err != nil { +		return "", err +	} +	respBody, err := ioutil.ReadAll(resp.Body) +	resp.Body.Close() +	if err != nil { +		return "", err +	} +	type RecordResp struct {  
Please use anonymous structs for a simpler code, similarly to these 5-10 lines
 
 
In pkg/update/update.go:
 > +	postValues.Set("sub_domain", host) +	req, err = http.NewRequest(http.MethodPost, "https://dnsapi.cn/Record.Ddns", bytes.NewBufferString(postValues.Encode())) +	if err != nil { +		return "", err +	} +	req.Header.Set("Content-Type", "application/x-www-form-urlencoded") +	resp, err = httpClient.Do(req) +	if err != nil { +		return "", err +	} +	respBody, err = ioutil.ReadAll(resp.Body) +	resp.Body.Close() +	if err != nil { +		return "", err +	} +	type DDNSResp struct {  
use anonynous strucrs here as well, thanks!
 
 
In pkg/update/update.go:
 > +	if err != nil { +		return "", err +	} +	type DDNSResp struct { +		Record struct { +			ID    int64  `json:"id"` +			Value string `json:"value"` +			Name  string `json:"name"` +		} `json:"record"` +	} +	ddnsResp := new(DDNSResp) +	err = json.Unmarshal(respBody, ddnsResp) +	if err != nil { +		return "", err +	} +	if ddnsResp.Record.Value != ip {  
Can your DNS provider determine your public IP automatically? Or do have to use another method to obtain your public IP (what seems to be the case from the code). If so, don't return the newIp and only check that the newIp is set to the wanted Ip (already done).

pkg/update/update_test.go
cnjack: I will delete the test file, the test file is just make sure the func is work,third party api is a dependence, I should make it is worked

cnjack:
this time should be good.

go.mod
```diff
@@ -3,6 +3,7 @@ module ddns-updater
 go 1.12
 
 require (
+	github.com/BurntSushi/toml v0.3.1 // indirect
```
qdm12: Why do you need `github.com/BurntSushi/toml`? I don't see it used anywhere I think.

pkg/update/update.go
```diff
@@ -485,26 +483,21 @@ func updateDnsPod(httpClient *http.Client, domain, host, token, ip string) (newI
 		return "", err
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
-	resp, err := httpClient.Do(req)
+	var content []byte
+	_, content, err = network.DoHTTPRequest(httpClient, req)
```
qdm12: Please do:

```go
status, content, err := network.DoHTTPRequest(httpClient, req)
if err != nil {
    return err
}
if status >= 300 {
   return "", fmt.Errorf("HTTP status %d", status)
}
```

as the HTTP status is important too.

pkg/update/update.go
```diff
@@ -485,26 +483,21 @@ func updateDnsPod(httpClient *http.Client, domain, host, token, ip string) (newI
...
 	if err != nil {
 		return "", err
 	}
-	respBody, err := ioutil.ReadAll(resp.Body)
-	resp.Body.Close()
-	if err != nil {
-		return "", err
-	}
-	type RecordResp struct {
+	recordResp := new(struct {
```
qdm12: just do 

```go
var recordResp struct {
    Records []struct {
        ID    string `json:"id"`
        Value string `json:"value"`
        Type  string `json:"type"`
        Name  string `json:"name"`
        Line  string `json:"line"`
    } `json:"records"`
}
```

The `new` keyboard is cumbersome and there is no need for pointers here `*`.

Adapt also the next line to 

```go
err = json.Unmarshal(content, &recordResp)
```

pkg/update/update.go
```diff
@@ -535,29 +528,23 @@ func updateDnsPod(httpClient *http.Client, domain, host, token, ip string) (newI
 		return "", err
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
-	resp, err = httpClient.Do(req)
+	_, content, err = network.DoHTTPRequest(httpClient, req)
```
qdm12: Same here, please check the HTTP status.

pkg/update/update.go
```diff
@@ -535,29 +528,23 @@ func updateDnsPod(httpClient *http.Client, domain, host, token, ip string) (newI
...
 	if err != nil {
 		return "", err
 	}
-	respBody, err = ioutil.ReadAll(resp.Body)
-	resp.Body.Close()
-	if err != nil {
-		return "", err
-	}
-	type DDNSResp struct {
+	ddnsResp := new(struct {
```
qdm12: Same here, please avoid `new` and use `var varname struct ...` directly.

pkg/update/update.go
```diff
@@ -535,29 +528,23 @@ func updateDnsPod(httpClient *http.Client, domain, host, token, ip string) (newI
...
+	err = json.Unmarshal(content, ddnsResp)
 	if err != nil {
 		return "", err
 	}
 	if ddnsResp.Record.Value != ip {
-		return "", errors.New("ip not set")
+		return "", fmt.Errorf("ip not set")
 	}
-	return ddnsResp.Record.Value, nil
+	return "", nil
```
qdm12: Please do not return an empty string, simply return the error. So change the function signature to return an error only and not a IP string and an error.

go.mod
cnjack: I just run `go mod tidy` and I don't known where i used, before i push the pr it is already in the go.mod

PR #30, opened 2020-02-16
Add support for Cloudflare API Tokens

mbellgb:
This PR adds support for authenticating using Cloudflare API tokens, rather than
having to use your personal API key.

This is a more secure option, since the API tokens can be rolled over a lot
easier, and can be locked down to specific resources and actions.

I've tested this working on my own machine and seems to work great!

[Information about the API tokens](https://blog.cloudflare.com/api-tokens-general-availability/)

Thanks for releasing this software, it's been really handy for my home setup!

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,7 @@ Please then refer to your specific DNS host provider in the section below for ev
     - Either:
         - Email `"email"` and Key `"key"`
         - User service key `"user_service_key"`
+        - API Token `"token"`, configured with DNS edit permissions for your DNS name's zone.
 - Optional:
     - `"proxied"` is a boolean to use the proxy services of Cloudflare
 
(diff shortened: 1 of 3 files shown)


mbellgb:
Haha, no problem! You're more than welcome to pull me in for future PRs too if you ever need a second set of eyes!

qdm12:
Thanks! I'll probably ask you for that refactoring I'm currently doing :smile: I'm also planning on adding support to update AAAA records for ipv6, but that will come later.

qdm12:
Although it's a big refactor, feel free to criticize the refactor 😄 

https://github.com/qdm12/ddns-updater/pull/32

Next plan is to move away from sqlite to a json based data file 👍

mbellgb:
will do! :D

PR #51, opened 2020-04-12
Updated Cloudflare documentation in README.md.

tools400:
I had a hard time figuring out which keys and identifiers I had to use with Cloudflare. One of the main problems was the dns record identifier. Obviously I was not smart enough to scroll down to the bottom of the page. There I would have found the missing piece of the puzzle, the Cloudflare "List DNS Records" API.
I slightly updated the README.md to make it easier for further users. Feel free to adopt the changes in case you like it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,12 +148,12 @@ Namecheap:
 
 Cloudflare:
 
-- `"zone_identifier"`
-- `"identifier"`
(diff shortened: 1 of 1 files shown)


tools400:
Sorry, I did not read your contributing guidelines beforehand.

Issue #100, opened 2020-09-18
docker reports unhealthy


KillerKelvUK:
Running as a docker container on my nas.  Functionality is spot-on with zero problems save docker reports the container as "unhealthy" 100% of the time.

Given the container is updating the couple of ddns entries I've supplied this is absolutely a minor niggle and "will live with", just wondering if I'm missing something simple is all?

Thanks,

K

qdm12:
Hello there,

The healthcheck ([code])(https://github.com/qdm12/ddns-updater/blob/master/internal/healthcheck/healthcheck.go#L23)) runs a DNS lookup over each of your records to check that the IP address it receives is the same as the one it has in its memory/json file. You should however have log lines starting with `unhealthy: ...` stating what's the error, do you have these? Thanks!

KillerKelvUK:
Hi, nothing stating unheathly, just one in the below which I'm guessing is a 1 time event...the containers been up 16 hours when the log below was gathered.

```
$ docker logs ddns-updater
=========================================
=========================================
========= DDNS Universal Updater ========
=========================================
=== Made with ❤️  by github.com/qdm12 ====
=========================================

Running version latest built on 2020-09-05T15:05:31Z (commit 00efca4)


🔧  Need help? https://github.com/qdm12/ddns-updater/issues/new
💻  Email? quentin.mcgaw@gmail.com
☕  Slack? Join from the Slack button on Github
💸  Help me? https://github.com/sponsors/qdm12
2020-09-07T08:34:31.245Z        ERROR   cannot get public ipv4 or ipv6 address: Get "https://domains.google.com/checkip": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
=========================================
=========================================
========= DDNS Universal Updater ========
=========================================
=== Made with ❤️  by github.com/qdm12 ====
=========================================

Running version latest built on 2020-09-05T15:05:31Z (commit 00efca4)


🔧  Need help? https://github.com/qdm12/ddns-updater/issues/new
💻  Email? quentin.mcgaw@gmail.com
☕  Slack? Join from the Slack button on Github
💸  Help me? https://github.com/sponsors/qdm12
=========================================
=========================================
========= DDNS Universal Updater ========
=========================================
=== Made with ❤️  by github.com/qdm12 ====
=========================================

Running version latest built on 2020-09-05T15:05:31Z (commit 00efca4)


🔧  Need help? https://github.com/qdm12/ddns-updater/issues/new
[...]

PR #103, opened 2020-09-24
Alpine

qdm12:
Alpine image, refers to #100


qdm12:
No longer relevant, this was used for debugging only.

PR #123, opened 2020-12-08
Google Domains instructions

gauravspatel:
I had to struggle through reverse engineering the /nic/update call and figure out why I kept getting nohost.  I discovered that I need to setup Dynamic DNS and also figured out how to do wildcards.  Adding to the README so others can save time.

Great image!  Thanks for the work.


gauravspatel:
I messed something up - resubmitting.

PR #124, opened 2020-12-08
 Google Domain - Dynamic DNS + Wildcards

gauravspatel:
I had to struggle through reverse engineering the /nic/update call and figure out why I kept getting nohost. I discovered that I need to setup Dynamic DNS and also figured out how to do wildcards. Adding to the README so others can save time.

Great image! Thanks for the work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,8 @@ DYNDNS:
 
 Google:
 
+Note that you must enable Dynamic DNS in the Synthetic records section of DNS management. The username and password is generated once you . You cannot create a wildcard entry for Dynamic DNS but you can CNAME `"*"` to the `"@"` domain for the same effect.
+
 - `"domain"`
 - `"host"` is your host and can be a subdomain or `"@"` or `"*"`
 - `"username"`


README.md
```diff
@@ -217,6 +217,8 @@ DYNDNS:
 
 Google:
 
+Note that you must enable Dynamic DNS in the Synthetic records section of DNS management. The username and password is generated once you . You cannot create a wildcard entry for Dynamic DNS but you can CNAME `"*"` to the `"@"` domain for the same effect.
```
qdm12, line 220: There is a missing piece after `is generated once you`.

qdm12, line 220: > You cannot create a wildcard entry for Dynamic DNS but you can CNAME `"*"` to the `"@"` domain for the same effect.

You mean that setting a CNAME record with key `"*"` and value `"@"` has the same effect right?

qdm12 (commented):
Thanks for the PR! Can you please just comment back on my comments to clarify a few things. I'll add it in the [Google Wiki section](https://github.com/qdm12/ddns-updater/wiki/Google) to keep the readme cleaner.

README.md
gauravspatel: The text should read:

Note that you must enable Dynamic DNS in the Synthetic records section of DNS management. The username and password is generated once you create the Dynamic DNS entry . You cannot create a wildcard entry for Dynamic DNS but you can CNAME `"*"` to the `"@"` domain for the same effect.

gauravspatel: Great catch...Yes, same effect.  I updated my text in the previous comment to help clarify.

qdm12:
I'll add your contribution in #128, closing this one for now. I mentioned you in the part you added thanks for the contribution!

PR #144, opened 2021-01-17
Linode support

### LINKED ISSUE #143: Feature request: Add Linode as a DNS provider
1. What's the feature?
Please add Linode as a DNS provider

2. Extra information?
- Within the Linode Cloud manager, create a `personalAccessToken` - it will be 64 chars. Assign it the `domain` permissions
- Obtain your `domainId` and `recordId` for the A record you want to update. I did this by inspecting the network tab as I manually updated the record. They will both be integers
- `PUT` request to update the record when your IP changes. Here is the JS to do so:

```
    const newIp = '127.0.0.1';
    const url = `https://api.linode.com/v4/domains/${LINODE_DOMAIN_ID}/records/${LINODE_RECORD_ID}`;
    const response = await fetch(url, {
      headers: {
        Authorization: `Bearer ${LINODE_PERSONAL_ACCESS_TOKEN}`,
        'Content-Type': 'application/json',
      },
      method: 'PUT',
      body: JSON.stringify({
        target: newIp,
      }),
    });
```

On success, a JSON response with the updated A record will be returned to you.

Thanks so much!

qdm12:
Fixes #143

### DIFF
--- a/internal/settings/linode.go
+++ b/internal/settings/linode.go
@@ -291,6 +291,7 @@ func (l *linode) createRecord(ctx context.Context, client *http.Client,
 	} else if !newIP.Equal(ip) {
 		return fmt.Errorf("%w: %s", ErrIPReceivedMismatch, newIP.String())
 	}
+	fmt.Println("DEBUG: Record created: ", responseData)
 
 	return nil
 }
--- a/internal/settings/linode.go
+++ b/internal/settings/linode.go
@@ -6,6 +6,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"net/url"
@@ -14,7 +16,6 @@ import (
 	"github.com/qdm12/ddns-updater/internal/constants"
 	"github.com/qdm12/ddns-updater/internal/models"
 	"github.com/qdm12/ddns-updater/internal/regex"
-	netlib "github.com/qdm12/golibs/network"
 )
 
 type linode struct {
@@ -85,7 +86,7 @@ func (l *linode) HTML() models.HTMLRow {
 }
 
 // Using https://www.linode.com/docs/api/domains/
-func (l *linode) Update(ctx context.Context, client netlib.Client, ip net.IP) (newIP net.IP, err error) {
+func (l *linode) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP net.IP, err error) {
 	domainID, err := l.getDomainID(ctx, client)
 	if err != nil {
 		return nil, fmt.Errorf("%w: %s", ErrGetDomainID, err)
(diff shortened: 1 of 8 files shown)


internal/settings/linode.go
```diff
@@ -0,0 +1,321 @@
...
+	return l.ipVersion
+}
+
+func (l *linode) BuildDomainName() string {
+	return buildDomainName(l.host, l.domain)
+}
+
+func (l *linode) HTML() models.HTMLRow {
+	return models.HTMLRow{
+		Domain:    models.HTML(fmt.Sprintf("<a href=\"http://%s\">%s</a>", l.BuildDomainName(), l.BuildDomainName())),
```
AnujRNair, line 80: The config json asks to add the domain name (example.com), but this line concatenates the host and domain name (subdomain.example.com) under the header "Domain" - is this intentional, given that we have an extra column for host directly after this one, or should the domain in this column match the domain configured in the config file?

internal/settings/linode.go
```diff
@@ -0,0 +1,321 @@
...
+		Path:   "/v4/domains",
+	}
+
+	r, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
+	if err != nil {
+		return 0, err
+	}
+	r.Header.Set("Content-Type", "application/json")
+	r.Header.Set("Authorization", "Bearer "+l.token)
+	r.Header.Set("oauth", "domains:read_only")
```
AnujRNair, line 135: I may be mistaken, but given we're using a personal access token, I don't think we need to define oauth headers?

qdm12: Hey thanks for the review 👍 

That's what they have on their documentation.
You can probably define tokens with different kind of permissions.
It might not work if you don't precise which permission you want, but I can't test it for sure, I don't have an account 😉

internal/settings/linode.go
qdm12: Yes this is intentional, because a particular row is about host + domain, and can only be about the domain only if the host is set to `@`. But you're right it's a tiny bit confusing, I'll take it into account when I do a v2 UI 👍

internal/settings/linode.go
AnujRNair: Sounds good! When creating the token, I had to assign permissions to it, but better to go with what the documentation says :)

internal/settings/linode.go
AnujRNair: Thanks!

internal/settings/linode.go
qdm12: Oh interesting which permissions? Are these the same as in the code? So I can add these to the Linode documentation.

AnujRNair: Go to setting and API Tokens:
Screen Shot 2021-01-17 at 6 56 51 PM

API Tokens Tab:
Screen Shot 2021-01-17 at 6 57 38 PM

Add Personal Access Token.
Screen Shot 2021-01-17 at 6 53 54 PM

I have a token with `domains` set, with read and write privileges, set to never expire

internal/settings/linode.go
```diff
@@ -0,0 +1,328 @@
...
+
+func (l *linode) isValid() error {
+	if len(l.token) == 0 {
+		return ErrEmptyToken
+	}
+	return nil
+}
+
+func (l *linode) String() string {
+	return toString(l.domain, l.host, constants.LUADNS, l.ipVersion)
```
AnujRNair, line 56: Should this be Linode?

qdm12: It definitely should 👍  Although that's just a display thing, it won't affect the rest of the code. Thanks for spotting it!

PR #149, opened 2021-01-25
Gandi support

fredericrous:
Add support for http://gandi.net/ registrar

Notes:

- this implementation uses gandiv5 API
- Gandi API defaults to a TTL of 10800 when no TTL is provided when we do a PUT request to update the record. I set the default TTL to 3600 in ddns-updater but it can be overwritten from the configuration

please let me know what you think @qdm12  

This PR resolves https://github.com/qdm12/ddns-updater/issues/150

### DIFF
--- a/docs/gandi.md
+++ b/docs/gandi.md
@@ -12,7 +12,7 @@ This provider uses Gandi v5 API
     {
       "provider": "gandi",
       "domain": "domain.com",
-      "name": "@",
+      "host": "@",
       "key": "key",
       "ttl": 3600,
       "ip_version": "ipv4",
@@ -21,12 +21,16 @@ This provider uses Gandi v5 API
 }
 ```
 
-If no ttl is defined, default is 3600
-
 ### Compulsory parameters
 
 - `"domain"` is your fqdn, for example `subdomain.duckdns.org`
+- `"host"`
 - `"key"`
+
+### Optional parameters
+
+- `"ip_version"` can be `ipv4` (A records) or `ipv6` (AAAA records), defaults to `ipv4 or ipv6`
+- `"ttl"` default is `3600`
 
 ## Domain setup
 
--- a/docs/gandi.md
+++ b/docs/gandi.md
@@ -1,0 +1,33 @@
+# Gandi
+
+This provider uses Gandi v5 API
+
+## Configuration
+
+### Example
+
+```json
+{
+  "settings": [
+    {
+      "provider": "gandi",
+      "domain": "domain.com",
+      "name": "@",
+      "key": "key",
+      "ttl": 3600,
+      "ip_version": "ipv4",
+    }
+  ]
+}
+```
+
+If no ttl is defined, default is 3600
+
+### Compulsory parameters
+
+- `"domain"` is your fqdn, for example `subdomain.duckdns.org`
+- `"key"`
+
+## Domain setup
+
+[Gandi Documentation Website](https://docs.gandi.net/en/domain_names/advanced_users/api.html#gandi-s-api)
--- a/docs/gandi.md
+++ b/docs/gandi.md
@@ -23,7 +23,7 @@ This provider uses Gandi v5 API
 
 ### Compulsory parameters
 
-- `"domain"` is your fqdn, for example `subdomain.duckdns.org`
+- `"domain"`
 - `"host"`
 - `"key"`
 
--- a/docs/gandi.md
+++ b/docs/gandi.md
@@ -24,7 +24,7 @@ This provider uses Gandi v5 API
 ### Compulsory parameters
 
 - `"domain"`
-- `"host"`
+- `"host"` which can be a subdomain, `@` or a wildcard `*`
 - `"key"`
 
 ### Optional parameters
--- a/internal/regex/regex.go
+++ b/internal/regex/regex.go
@@ -3,6 +3,7 @@ package regex
 import "regexp"
 
 type Matcher interface {
+	GandiKey(s string) bool
 	GodaddyKey(s string) bool
 	GodaddySecret(s string) bool
 	DuckDNSToken(s string) bool
@@ -15,12 +16,16 @@ type Matcher interface {
 }
 
 type matcher struct {
-	goDaddyKey, goDaddySecret, duckDNSToken, namecheapPassword, dreamhostKey, cloudflareKey, cloudflareUserServiceKey,
-	dnsOMaticUsername, dnsOMaticPassword *regexp.Regexp
+	gandiKey, goDaddyKey, goDaddySecret, duckDNSToken, namecheapPassword, dreamhostKey, cloudflareKey,
+	cloudflareUserServiceKey, dnsOMaticUsername, dnsOMaticPassword *regexp.Regexp
 }
 
 func NewMatcher() (m Matcher, err error) {
 	matcher := &matcher{}
+	matcher.gandiKey, err = regexp.Compile(`^[A-Za-z0-9]{24}$`)
+	if err != nil {
+		return nil, err
+	}
 	matcher.goDaddyKey, err = regexp.Compile(`^[A-Za-z0-9]{8,14}\_[A-Za-z0-9]{22}$`)
 	if err != nil {
 		return nil, err
@@ -60,6 +65,7 @@ func NewMatcher() (m Matcher, err error) {
 	return matcher, nil
 }
 
+func (m *matcher) GandiKey(s string) bool          { return m.gandiKey.MatchString(s) }
 func (m *matcher) GodaddyKey(s string) bool        { return m.goDaddyKey.MatchString(s) }
 func (m *matcher) GodaddySecret(s string) bool     { return m.goDaddySecret.MatchString(s) }
 func (m *matcher) DuckDNSToken(s string) bool      { return m.duckDNSToken.MatchString(s) }
--- a/internal/regex/regex.go
+++ b/internal/regex/regex.go
@@ -16,8 +16,8 @@ type Matcher interface {
 }
 
 type matcher struct {
-	gandiKey, goDaddyKey, goDaddySecret, duckDNSToken, namecheapPassword, dreamhostKey, cloudflareKey,
-	cloudflareUserServiceKey, dnsOMaticUsername, dnsOMaticPassword *regexp.Regexp
+	goDaddyKey, goDaddySecret, duckDNSToken, namecheapPassword, dreamhostKey, cloudflareKey,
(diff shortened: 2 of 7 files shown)


fredericrous:
Example of config that works for my use-case:

```json
{
    "settings": [
        {
            "provider": "gandi",
            "domain": "mydomain.com",
            "host": "@",
            "name": "@",
            "key": "mysecretkey",
            "ip_version": "ipv4"
        },
        {
            "provider": "gandi",
            "domain": "mydomain.com",
            "host": "@",
            "name": "*",
            "key": "mysecretkey",
            "ip_version": "ipv4"
        }
    ]
}
```

updates my DNS record as the following

image

fredericrous:
The tests look good to me, I don't think the action "Run docker/login-action" fails because of this PR

docs/gandi.md
```diff
@@ -0,0 +1,37 @@
...
+      "ttl": 3600,
+      "ip_version": "ipv4",
+    }
+  ]
+}
+```
+
+### Compulsory parameters
+
+- `"domain"` is your fqdn, for example `subdomain.duckdns.org`
```
qdm12, line 26:
```suggestion
- `"domain"`
```

docs/gandi.md
```diff
@@ -0,0 +1,37 @@
...
+- `"host"`
```
qdm12, line 27:
```suggestion
- `"host"` which can be a subdomain, `@` or a wildcard `*`
```

internal/regex/regex.go
```diff
@@ -15,12 +16,16 @@ type Matcher interface {
 }
 
 type matcher struct {
-	goDaddyKey, goDaddySecret, duckDNSToken, namecheapPassword, dreamhostKey, cloudflareKey, cloudflareUserServiceKey,
-	dnsOMaticUsername, dnsOMaticPassword *regexp.Regexp
+	gandiKey, goDaddyKey, goDaddySecret, duckDNSToken, namecheapPassword, dreamhostKey, cloudflareKey,
+	cloudflareUserServiceKey, dnsOMaticUsername, dnsOMaticPassword *regexp.Regexp
```
qdm12, line 20: Just being a low git diff maniac 😄 

```suggestion
	goDaddyKey, goDaddySecret, duckDNSToken, namecheapPassword, dreamhostKey, cloudflareKey,
	cloudflareUserServiceKey, dnsOMaticUsername, dnsOMaticPassword, gandiKey *regexp.Regexp
```

internal/settings/gandi.go
```diff
@@ -0,0 +1,199 @@
...
+
+	recordIP, err := g.getRecordIP(ctx, recordType, client)
+	if err != nil && err != ErrNoResultReceived { // if no ip was defined before, let's proceed with the update
+		return nil, fmt.Errorf("%s: %w", ErrGetRecordIP, err)
+	}
+
+	oldIP := net.ParseIP(recordIP)
+	if ip.Equal(oldIP) { // success, nothing to change
+		return ip, nil
+	}
```
qdm12, line 154: Sorry for the wasted effort, but this can be removed as it's already doing a DNS resolution against your domain to check if the IP changed in some other part of the code. **Except** if the domain can be proxied (which makes the DNS resolution mechanism check not working), but for now I think only Cloudflare offers such service.

```suggestion
```

internal/settings/gandi.go
```diff
@@ -0,0 +1,199 @@
...
+		Path:   fmt.Sprintf("/api/v5/domains/%s/records/%s/%s", g.domain, g.host, recordType),
+	}
+
+	buffer := bytes.NewBuffer(nil)
+	encoder := json.NewEncoder(buffer)
+	requestData := struct {
+		Values [1]string `json:"rrset_values"`
+		TTL    int       `json:"rrset_ttl"`
+	}{
+		Values: [1]string{ip.To4().String()},
```
qdm12, line 168: How about ipv6? Does gandi support ipv6 addresses (AAAA records)?

internal/settings/gandi.go
```diff
@@ -0,0 +1,199 @@
...
+		TTL: func() int {
+			ttl := DefaultTTL
+			if g.ttl != 0 {
+				ttl = g.ttl
+			}
+			return ttl
+		}(),
```
qdm12, line 175: this is fancy but please just put it above the `requestData := ...` as:

```go
const defaultTTL = 3600
ttl := defaultTTL
if g.ttl != 0 {
	ttl = g.ttl
}
```

to keep it simple, thanks!

internal/settings/gandi.go
```diff
@@ -0,0 +1,199 @@
...
+)
+
+const DefaultTTL = 3600
+
+type gandi struct {
+	domain    string
+	host      string
+	ttl       int
+	ipVersion models.IPVersion
+	dnsLookup bool
```
qdm12, line 24: No need for the `dnsLookup` option, only `cloudflare` needs it actually because it proxies domains

```suggestion
```

internal/settings/gandi.go
```diff
@@ -0,0 +1,199 @@
...
+	key       string
+}
+
+func NewGandi(data json.RawMessage, domain, host string, ipVersion models.IPVersion,
+	noDNSLookup bool, matcher regex.Matcher) (s Settings, err error) {
```
qdm12, line 29:
```suggestion
	_ bool, matcher regex.Matcher) (s Settings, err error) {
```

internal/settings/gandi.go
```diff
@@ -0,0 +1,199 @@
...
+		Key string `json:"key"`
+		TTL int    `json:"ttl"`
+	}{}
+	if err := json.Unmarshal(data, &extraSettings); err != nil {
+		return nil, err
+	}
+	d := &gandi{
+		domain:    domain,
+		host:      host,
+		ipVersion: ipVersion,
```
qdm12, line 40:
```suggestion
```

internal/settings/gandi.go
```diff
@@ -0,0 +1,199 @@
...
+func (g *gandi) Domain() string {
+	return g.domain
+}
+
+func (g *gandi) Host() string {
+	return g.host
+}
+
+func (g *gandi) DNSLookup() bool {
+	return g.dnsLookup
```
qdm12, line 71:
```suggestion
	return true
```

internal/settings/gandi.go
```diff
@@ -0,0 +1,199 @@
...
+	"net"
+	"net/http"
+	"net/url"
+
+	"github.com/qdm12/ddns-updater/internal/constants"
+	"github.com/qdm12/ddns-updater/internal/models"
+	"github.com/qdm12/ddns-updater/internal/regex"
+)
+
+const DefaultTTL = 3600
```
qdm12, line 17: Better to inline the constant where it's needed (see suggestion below)

```suggestion
```

qdm12 (commented):
Thanks for the contribution 💯  A few comments and suggestions, but essentially we need to figure out:

- how to handle ipv6, if Gandi supports AAAA records
- remove the getrecordip as I believe it's not necessary

qdm12:
I also fixed the CI for external pull request (I hope!), so feel free to rebase it on the master branch of my repo if you want. Thanks!

internal/settings/gandi.go
fredericrous: Here I was trying to check if the IP on the registrar is correct or not in order to decide if we want to update the IP or not. Indeed it might not be worth checking, it duplicates the number of calls for no gain

internal/settings/gandi.go
fredericrous: indeed gandi supports ipv6

fredericrous: I implemented logic to either convert to ipv6 or ipv4, let me know what you think

fredericrous:
I tested with the changes we did and now the Update method seemed to not be called anymore, it's pretty late here in Ireland, I'll take a look tomorrow

internal/settings/errors.go
```diff
@@ -31,6 +32,7 @@ var (
 	ErrCreateRecord    = errors.New("cannot create record")
 	ErrGetDomainID     = errors.New("cannot get domain ID")
 	ErrGetRecordID     = errors.New("cannot get record ID")
+	ErrGetRecordIP     = errors.New("cannot get record IP")
```
qdm12, line 35: 😉 

```suggestion
```

internal/settings/gandi.go
```diff
@@ -0,0 +1,147 @@
...
+	}{
+		Values: [1]string{ipStr},
+		TTL: func() int {
+			ttl := 3600
+			if g.ttl != 0 {
+				ttl = g.ttl
+			}
+			return ttl
+		}(),
+	}
```
qdm12, line 124: To keep it simple 😉 

```suggestion
	const defaultTTL = 3600
	ttl := defaultTTL
	if g.ttl != 0 {
		ttl = g.ttl
	}
	requestData := struct {
		Values [1]string `json:"rrset_values"`
		TTL    int       `json:"rrset_ttl"`
	}{
		Values: [1]string{ipStr},
		TTL: ttl,
	}
```

internal/settings/gandi.go
```diff
@@ -0,0 +1,147 @@
...
+	g.setHeaders(request)
+
+	response, err := client.Do(request)
+	if err != nil {
+		return nil, err
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusCreated {
```
qdm12, line 141: Do we always get a status created, even if we update the record? 🤔

qdm12 (commented):
Nice 🥇 Looking good, just a few minor comments/questions.

qdm12:
> I tested with the changes we did and now the Update method seems to not be called anymore, it's pretty late here in Ireland, I'll take a look tomorrow

The update is only called if the DNS record mismatches your public IP address. If it's the same, it doesn't do an update.

You can try setting the IP address to `127.0.0.1` in the Gandi UI, and wait to check if it updates it.
You can also try with `nslookup yourdomain.com` to check when your Gandi web UI changes for `127.0.0.1` have propagated.

PS: It seems the CI fix I pushed fixed the CI for you, nice 👍

internal/settings/gandi.go
fredericrous: yes, I agree that is weird but I didn't design this API 😆

fredericrous:
Thank you for the tips @qdm12, no need to wait for propagation, it is a separate thing. I got the update working by setting DNSLookup to false, see commit f39a435

qdm12 (approved):

Issue #153, opened 2021-01-27
Feature request: Add support for *.yourdomain.tld on OVH


fredericrous:
1. What's the feature?

Currently, the OVH provider updates the IP with OVH's dynhost service. This does not allow to update * records.

I am using Traefik, all the Softwares I use are under subdomains. I need to update A @ record + A * record

2. Extra information?

I created the following PR: https://github.com/qdm12/ddns-updater/pull/154

qdm12:
Thanks! I invited you as a collaborator, that should make it easier for you to do PRs and all 😉

PR #154, opened 2021-01-27
[ovh] Add support for * sub domain

fredericrous:
This PR introduces support for subdomain record actualization (A * or else)

This solves issue https://github.com/qdm12/ddns-updater/issues/153

This PR introduces a new go package, ovh-go. I am usually reluctant to add external packages. However, the requests to OVH API have to be signed with a complicated algorithm. Not using their packages would mean maintaining kinda the same amount of code the ovh-go package has + introduce tests for the request signature which are already done in the ovh-go package.

I tested record creation and record update for A @ and A *. The PR also supports AAAA but I don't use it


docs/ovh.md
```diff
@@ -24,12 +24,33 @@
...
+- `"app_secret"`
+- `"consumer_key"`
+
+The ZoneDNS implementation allows you to update any record name including *.yourdomain.tld
+
 ### Optional parameters
 
 - `"ip_version"` can be `ipv4` (A records) or `ipv6` (AAAA records), defaults to `ipv4 or ipv6`
 - `"provider_ip"` can be set to `true` to let your DNS provider determine your IPv4 address (and/or IPv6 address) automatically when you send an update request, without sending the new IP address detected by the program in the request.
+- `"zone_dns"` select between two modes, OVH's dynamic hosting service (`false`) or OVH's API (`true`). Default is `false`
```
qdm12, line 46: Maybe make that a string in case in the future there is another mode? As in, it's unlikely, but it's good practice 😉 Plus it's easier to remember for the user to have `"mode": "dynamic"` or `"mode": "api"` or something like that, than remembering what `true` means.

docs/ovh.md
```diff
@@ -24,12 +24,33 @@
...
 
 ## Domain setup
+
+If you are using DynHost:
+
+https://docs.ovh.com/ie/en/domains/hosting_dynhost/
+
+If you are using the ZoneDNS api:
+
+https://docs.ovh.com/gb/en/customer/first-steps-with-ovh-api/
```
qdm12, line 56:
```suggestion
- If you use DynHost: [docs.ovh.com/ie/en/domains/hosting_dynhost](https://docs.ovh.com/ie/en/domains/hosting_dynhost/)
- If you use the ZoneDNS API: [docs.ovh.com/gb/en/customer/first-steps-with-ovh-api](https://docs.ovh.com/gb/en/customer/first-steps-with-ovh-api/)
```

go.sum
```diff
@@ -44,8 +44,10 @@ github.com/gomodule/redigo v2.0.0+incompatible/go.mod h1:B4C85qUVwatsJoIUNIfCRsp
 github.com/google/renameio v0.1.0/go.mod h1:KWCgfxg9yswjAJkECMjeO8J8rahYeXnNhOm40UhjYkI=
 github.com/google/uuid v1.0.0 h1:b4Gk+7WdP/d3HZH8EJsZpvV7EtDOgaZLtnaNGIu1adA=
 github.com/google/uuid v1.0.0/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
+github.com/gopherjs/gopherjs v0.0.0-20181017120253-0766667cb4d1/go.mod h1:wJfORRmW1u3UXTncJ5qlYoELFm8eSnnEO6hX4iZ3EWY=
```
qdm12, line 47: Strange strange strange, but fine I guess 😅

internal/settings/ovh.go
```diff
@@ -78,7 +105,7 @@ func (o *ovh) IPVersion() models.IPVersion {
 }
 
 func (o *ovh) DNSLookup() bool {
-	return o.dnsLookup
+	return false
```
qdm12, line 108: Oh god this is just horrible, I need to fix it 🤣

internal/settings/ovh.go
```diff
@@ -142,3 +169,77 @@ func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP
...
+		recordType = AAAA
+		ipStr = ip.To16().String()
+	} else {
+		ipStr = ip.To4().String()
+	}
+	// subDomain filter of the ovh api expect an empty string to get @ record
+	subDomain := o.host
+	if subDomain == "@" {
+		subDomain = ""
+	}
```
qdm12, line 186: How about `*`? 🤔

internal/settings/ovh.go
```diff
@@ -142,3 +169,77 @@ func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP
...
+	// get existing records
+	var recordIDs []uint64
+	if err := client.Get(fmt.Sprintf("/domain/zone/%s/record?fieldType=%s&subDomain=%s", o.domain,
+		recordType, subDomain), &recordIDs); err != nil {
+		return ip, err
+	}
```
qdm12, line 192: I'm going to sound like a picky coder, but can you please use `client.Do` instead and create the request with `request := http.NewRequestWithContext(ctx, http.MethodGet, "url", nil)`? That way the request/client processing stops if the context gets canceled, I think in this case it only happens when the program gets interrupted but it could be important.

Same comment for the `client.Post`. It's a bit longer but there is no way around it unfortunately for now with the Go API. Note that for the Post, you can use

```go
buffer := bytes.NewBuffer(nil)
encoder := json.NewEncoder(buffer)
if err := json.Encode(yourstruct); err != nil {
     ....
}
```

to encode your json struct and then pass the buffer as the last argument of `http.NewRequestWithContext`

internal/settings/ovh.go
```diff
@@ -142,3 +169,77 @@ func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP
...
+	// subDomain filter of the ovh api expect an empty string to get @ record
+	subDomain := o.host
+	if subDomain == "@" {
+		subDomain = ""
+	}
+	// get existing records
+	var recordIDs []uint64
+	if err := client.Get(fmt.Sprintf("/domain/zone/%s/record?fieldType=%s&subDomain=%s", o.domain,
+		recordType, subDomain), &recordIDs); err != nil {
+		return ip, err
```
qdm12, line 191:
```suggestion
		return nil, err
```

just for clarity, the newIP is `nil` because we encountered an error. The comment applies for the few returns below.

internal/settings/ovh.go
```diff
@@ -142,3 +169,77 @@ func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP
...
+			o.apiEndpoint,
+			o.appKey,
+			o.appSecret,
+			o.consumerKey,
+		)
+		newIP, err = o.updateWithZoneDNS(ovhClientInstance, ip)
+	} else {
+		newIP, err = o.updateWithDynHost(ctx, client, ip)
+	}
+	return newIP, err
```
qdm12, line 244: Better to return early to have less indentation

```suggestion
	if !o.zoneDNS {
		return  o.updateWithDynHost(ctx, client, ip)
	}
	if len(o.apiEndpoint) == 0 {	
		o.apiEndpoint = "ovh-eu"
	}
	ovhClientInstance, _ := ovhClient.NewClient(
		o.apiEndpoint,
		o.appKey,
		o.appSecret,
		o.consumerKey,
	)
	return o.updateWithZoneDNS(ovhClientInstance, ip)
```

internal/settings/ovh.go
```diff
@@ -142,3 +169,77 @@ func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP
...
+
+	return ip, nil
+}
+
+func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP net.IP, err error) {
+	if o.zoneDNS {
+		if len(o.apiEndpoint) == 0 {
+			o.apiEndpoint = "ovh-eu"
+		}
+		ovhClientInstance, _ := ovhClient.NewClient(
```
qdm12, line 234: if that's an error returned `_` let's check it

internal/settings/ovh.go
```diff
@@ -142,3 +169,77 @@ func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP
...
+	}
+
+	return ip, nil
+}
+
+func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP net.IP, err error) {
+	if o.zoneDNS {
+		if len(o.apiEndpoint) == 0 {
+			o.apiEndpoint = "ovh-eu"
+		}
```
qdm12, line 233: Being picky here, but for futureproofing (in case we can live change settings with an UI etc.), do not modify the fields of the struct here so instead have something like

```suggestion
		const defaultEndpoint = "ovh-eu"
		apiEndpoint := defaultEndpoint
		if len(o.apiEndpoint) > 0 {
			apiEndpoint = o.apiEndpoint
		}
```

And then use `apiEndpoint` instead of `o.apiEndpoint`.

OR maybe simpler, just set the default when it creates the `NewOVH`, that way it won't get modified every time an update is made. Although that brings the problem that modifying the state and changing `o.apiEndpoint` to `""` will make it still `""` and not `ovh-eu`

qdm12 (commented):
Nice thanks! 👍 A few comments here and there 😉

docs/ovh.md
fredericrous: That's a very valid point ^^ I'll implement that

internal/settings/ovh.go
fredericrous: star `*` works, I don't know why `@` doesn't. Instead, we have to pass an empty string to `&subdomain=` argument. And if we do not supply this argument there is no filter (except fieldType) and every records (except those filtered out by fieldType) are returned. So basically the default I coded is returned `@` unless another subdomain (`host`) is set

internal/settings/ovh.go
fredericrous: client.Do is used under the hood https://github.com/ovh/go-ovh/blob/302dfc012714871f55e4ce3bae69f903fe28327b/ovh/ovh.go#L387

go.sum
fredericrous: this comes from the test library used by the go-ovh package github.com/smartystreets/goconvey

fredericrous:
Thank you for the great review @qdm12  I think I implemented everything. I tested and this still works 🥳

internal/settings/ovh.go
qdm12: Oh my bad, I thought `client` was an `*http.Client`, glad you found these methods too! Perfect!

internal/settings/ovh.go
```diff
@@ -142,3 +169,81 @@ func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP
...
+	}
+	// subDomain filter of the ovh api expect an empty string to get @ record
+	subDomain := o.host
+	if subDomain == "@" {
+		subDomain = ""
+	}
+	// get existing records
+	var recordIDs []uint64
+	if err := client.GetWithContext(ctx, fmt.Sprintf("/domain/zone/%s/record?fieldType=%s&subDomain=%s", o.domain,
+		recordType, subDomain), &recordIDs); err != nil {
```
qdm12, line 190: nitpicking

```suggestion
	url := fmt.Sprintf("/domain/zone/%s/record?fieldType=%s&subDomain=%s", o.domain, recordType, subDomain)
	if err := client.GetWithContext(ctx, url, &recordIDs); err != nil {
```

internal/settings/ovh.go
```diff
@@ -142,3 +169,81 @@ func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP
...
+			FieldType string `json:"fieldType"`
+			SubDomain string `json:"subDomain"`
+			Target    string `json:"target"`
+		}{
+			FieldType: recordType,
+			SubDomain: subDomain,
+			Target:    ipStr,
+		}
+		if err := client.PostWithContext(ctx, fmt.Sprintf("/domain/zone/%s/record", o.domain),
+			&postRecordsParams, nil); err != nil {
```
qdm12, line 205:
```suggestion
		url := fmt.Sprintf("/domain/zone/%s/record", o.domain)
		if err := client.PostWithContext(ctx, url, &postRecordsParams, nil); err != nil {
```

internal/settings/ovh.go
```diff
@@ -142,3 +169,81 @@ func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP
...
+	} else {
+		// update existing record
+		putRecordsParams := struct {
+			Target string `json:"target"`
+		}{
+			Target: ipStr,
+		}
+		for _, recordID := range recordIDs {
+			if err := client.PutWithContext(ctx, fmt.Sprintf("/domain/zone/%s/record/%d", o.domain, recordID),
+				&putRecordsParams, nil); err != nil {
```
qdm12, line 217:
```suggestion
			url := fmt.Sprintf("/domain/zone/%s/record/%d", o.domain, recordID)
			if err := client.PutWithContext(ctx, url, &putRecordsParams, nil); err != nil {
```

internal/settings/ovh.go
```diff
@@ -142,3 +169,81 @@ func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP
...
+				return nil, err
+			}
+		}
+	}
+
+	if err := client.PostWithContext(ctx, fmt.Sprintf("/domain/zone/%s/refresh", o.domain), nil, nil); err != nil {
```
qdm12, line 223:
```suggestion
	url := fmt.Sprintf("/domain/zone/%s/refresh", o.domain)
	if err := client.PostWithContext(ctx, url, nil, nil); err != nil {
```

Also what's that call doing out of curiosity? 🧐

internal/settings/ovh.go
```diff
@@ -142,3 +169,81 @@ func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP
...
+	}
+	ovhClientInstance, err := ovhClient.NewClient(
+		apiEndpoint,
+		o.appKey,
+		o.appSecret,
+		o.consumerKey,
+	)
+	if err != nil {
+		return nil, err
+	}
```
qdm12, line 247: Nice 👍

internal/settings/ovh.go
```diff
@@ -142,3 +169,81 @@ func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP
...
+func (o *ovh) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP net.IP, err error) {
+	if o.mode != "api" {
+		return o.updateWithDynHost(ctx, client, ip)
+	}
+	const defaultEndpoint = "ovh-eu"
+	apiEndpoint := defaultEndpoint
+	if len(o.apiEndpoint) > 0 {
+		apiEndpoint = o.apiEndpoint
+	}
+	ovhClientInstance, err := ovhClient.NewClient(
```
qdm12, line 239: I'm wondering, is their client an interface by any chance? Or just a `struct`?

qdm12 (approved):
Just a few additional comments here and there.

Great job otherwise!

internal/settings/ovh.go
fredericrous: if we don't refresh, ovh doesn't validate/propagate the change

fredericrous:
I implemented these last changes, I tested and everything went ok 🎉

qdm12 (approved):

qdm12:
Merging because I really want to change that `DNSLookup()` method name right now 😄

crocobc:
Hi there,

I'm sorry if my post is not at the right place...

I tried to use API mode (with OVH)  few days ago but i'm having this bug [#319](https://github.com/qdm12/ddns-updater/issues/319)

So i tried with dynamic mode => it's working but not for @ domain.

I have "Failure (unknown response received: nohost )" error.

It seems to be related to this post ?

Thank you in advance for information.

Regards

PR #160, opened 2021-01-29
Remove no_dns_lookup and replace DNSLookup() method with Proxied()

qdm12:
This should only be used by Cloudflare when the record is proxied.

There is not much utility for the end user to disable this DNS lookup as it should be "free" to do (fast, low data, etc.) and adds a layer of check (if the user modifies the record in the cloudflare web ui, there is no way for the program to know it should update it - except making a call to cloudflare's API which is not implemented and probably rate limited).


fredericrous:
could it be possible to make the Proxied method optional to implement?

qdm12:
We could do a type assertion on the interface and check if it's Cloudflare and if it's proxied as it is really specific to cloudflare indeed. Not the prettiest but it might be cleaner. Will do that and we can see how it looks like.

fredericrous:
it sure would work and might be the simpler solution, however it feels hacky. Another solution I think of would be to compose the interface with a struct that implements the method with a default return value and let the providers implement this struct instead of the interface?

qdm12:
I tried both ways (doing a type assertion and through interface composition with a default method), and they both are horribly cognitively intense. I'll just merge it as is for now 😉

PR #159, opened 2021-01-28
ddnsgopher favicon

fredericrous:
Here is a favicon proposal. svg source + ico export

image

I also added the favicon link to the index.html most browsers knows where to find the favicon.ico. However, the recommended way by the w3c is to use the link tag.

Issue #75, opened 2020-05-31
Feature request: API HTTP server


qdm12:
1. What's the feature?

Setup a simple HTTP server to serve information and modify settings

2. Why do you need this feature?

Such that it can be used as an API consumed by the web ui react frontend

3. Extra information?

fredericrous:
I did a quick look online and the Iris framework seems to be a good choice for an API https://github.com/kataras/iris
It is actively maintained, very few issues are opened, there are loads of up to date examples. It looks easy to use/implement

qdm12:
Ehh sorry to disappoint, but I'd like to stay away from frameworks. Especially we just need I think a GET, POST and PUT endpoints and call it a day 👍 I'm fine with adding a (good, idiomatic, close to the Go standard library) router like [go-chi](https://github.com/go-chi/chi) though.

I'm very used to write HTTP servers in Go, what I can do is lay some foundations and then you can fill it up? But that will probably have to wait until the weekend 😄

fredericrous:
I was suggesting Iris because it looks super easy and readable from what I read from the section "Simple Handler" of their Readme.md . Also, I believe the go compiler removes unused code so using a framework shouldn't introduce too much heaviness. Anyway, I'm afraid to write an API in Go without structure, that's the main reason I suggested a framework. If you lay down some foundation I sure will be happy to help

qdm12:
> Anyway, I'm afraid to write an API in Go without structure

Totally understandable. It took me a while to figure a good structure, and I constantly see oddities in how HTTP servers are implemented / non testable. So it is not trivial indeed 😄

Right now I'm in the middle of solving #136 and will get to the HTTP server right after.

jdevera:
I came to suggest a similar feature. In my case, I would only need read support on such API. My goal is to show a summary in a [Homepage](https://gethomepage.dev/) widget so I can see if everything is fine or if something needs my attention.

Perhaps this could come as a first iteration before considering config changes 👼 .

Robonau:
even a simple json version of what is already on the current page would be fine.
just 
```json
[
  {
    "domain": "$domain",
    "owner": "$owner",
    "provider": "$provider",
    "iPVersion": "$iPVersion",
    "updateStatus": "$updateStatus",
    "timeSinceUpdate": "$time in updateStatus in unix time", 
    "currentIPs": "$currentIPs",
    "previousIPs": ["$previousIP","$previousIP"]
  }
]
```
on /api
would be fine

PR #164, opened 2021-02-03
HTTP server refactor

### LINKED ISSUE #148: Feature request: Fix HTTP route for favicon.ico
1. What's the feature?

2. Extra information?

qdm12:
- Fixes #148
- Paves way for #75 
- Unblocks #159 
- Also makes the call to `GET /update` blocking until all updates are done

I skipped adding unit tests as I would like to get moving with an API server and a client side UI (react) which would change the routes quite a bit (in a subsequent PR). Also sorry for the delay @fredericrous was caught up in other repos essentially 😄


fredericrous:
I'll test that now

fredericrous:
The server hangs

```
ddns-updater on  http-server [$!?] via 🐹 v1.15.7 ➜  LISTENING_PORT=4242 ./app                                                                                17:19:45
=========================================
=========================================
========= DDNS Universal Updater ========
=========================================
=== Made with ❤️  by github.com/qdm12 ====
=========================================

Running version unknown built on an unknown date (commit unknown)


🔧  Need help? https://github.com/qdm12/ddns-updater/issues/new
💻  Email? quentin.mcgaw@gmail.com
☕  Slack? Join from the Slack button on Github
💸  Help me? https://github.com/sponsors/qdm12
2021-02-07T17:19:46.684Z        INFO    Found 2 settings to update records
2021-02-07T17:19:46.988Z        INFO    Reading history from database: domain mydomain.com host @
2021-02-07T17:19:46.988Z        INFO    Reading history from database: domain mydomain.com host *
```

It should continue and display the following lines but doesn't

```
2021-02-07T16:29:54.670Z        INFO    healthcheck server: listening on 127.0.0.1:9999
2021-02-07T16:29:54.670Z        INFO    http server: listening on 0.0.0.0:8000
```

The issue seems to come from this line, or from the update Necessary function

`r.forceResult <- r.updateNecessary(ctx)`

If I print something before this line it displays, but after it doesn't.

From what I understand of Go, the channel should be closed in the necessary function and the error should be returned as a string instead of a stream, am I right?

cmd/updater/main.go
```diff
@@ -150,9 +150,9 @@ func _main(ctx context.Context, timeNow func() time.Time) int {
 	runner := update.NewRunner(db, updater, ipGetter, p.cooldown, logger, timeNow)
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
-	forceUpdate := make(chan struct{})
-	go runner.Run(ctx, p.period, forceUpdate)
-	forceUpdate <- struct{}{}
+
+	go runner.Run(ctx, p.period)
+	runner.ForceUpdate()
```
fredericrous, line 155: this hangs, should we replace this by something like `go runner.ForceUpdate(errors)` ? and check that error callback?

internal/server/error.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+type errorsJSONWrapper struct {
+	Errors []string `json:"errors"`
+}
+
+func httpErrors(w http.ResponseWriter, status int, errors []error) {
+	w.WriteHeader(status)
+
+	errs := make([]string, len(errors))
```
fredericrous, line 28: if you make() you need to close() somewhere, right?

internal/server/update.go
```diff
@@ -0,0 +1,18 @@
+package server
+
+import (
+	"net/http"
+)
+
+func (h *handlers) update(w http.ResponseWriter, r *http.Request) {
+	start := h.timeNow()
+	errors := h.runner.ForceUpdate()
```
fredericrous, line 9: issue here as well

fredericrous (changes requested):

internal/server/error.go
qdm12: Ha that's a fair question! You don't need to close slices (=dynamic arrays in Go) in Go, they just get garbage collected.

You can however close *channels* (like `make(chan string)`) but that's more of a developer thing rather than a performance/memory thing. Someone explained it better than me though:

> You needn't close every channel when you've finished with it. It's only necessary to close a channel when it is important to tell the receiving goroutines that all data have been sent. A channel that the garbage collector determinies to be unreachable will have its resources reclaimed whether or not it is closed. (Don't confuse this with the close operation for open files. It is important to call the Close method on every file when you've finished with it.)

cmd/updater/main.go
qdm12: That was because I forgot to create the `forceResult` channel, it's now fixed 😄 Thanks for finding this out!

And it's now running in a goroutine as we indeed don't want to block the rest of the goroutines with the first round of updates, especially since errors are logged within the updater so we don't need to collect errors.

fredericrous: hahaha, I should have checked the constructor ^^. I'm glad I could help

internal/server/error.go
fredericrous: thank you for taking the time to explain

fredericrous (approved):
Tested with success. However, I would expect /update to update the IP no matter what. And that's not the case with the current implementation. Indeed, if the local database contains the same IP already, it won't update the record on the provider website. It's an edge case but if I execute ddns-updater on my laptop. It updates the dns zone to my home's IP. I then stop ddns-updater because my test id done. Hours later I go on my provider's zonedns, the IP is still the one of my laptop and not the one of my NAS which is at my parent's (bless them ❤️). I decide to update the IP with a force update. I send an email to mum, ask her to click on the link http://ddns-updater.test/update and nothing happens. That's an example backed with a little story, for real I have several ways of knowing my IP without external help (other domain, saas app), but I think you get the use case?

I also created a PR here: https://github.com/qdm12/golibs/pull/2 I would love to get debug logs on functions getRecordIDsToUpdate, shouldUpdateRecordWithLookup, shouldUpdateRecordNoLookup

PS: favicon issue is fixed!

qdm12:
> Indeed, if the local database contains the same IP already, it won't update the record on the provider website

That's only the case for cloudflare if the record is proxied. Otherwise it DNS resolves your hostname and check it matches your current public IP address. [This](https://github.com/qdm12/ddns-updater/blob/master/internal/update/run.go#L146) is the codepath ran when there is an update call (through http or with the timer). Are you sure it doesn't update it?

Let's merge this PR for now, as well as the favicon one, but let's continue the discussion here 👍 

Regarding golibs, thanks for that contribution, it's merged now! And yes adding some debug logs could be nice.

On a side note, I'm actually reworking the logger implementation (maybe interface) to have something [like this](https://github.com/WireGuard/wireguard-go/blob/master/device/logger.go#L36) and ideally have different implementations such as zap & glog.

fredericrous:
Yeah it didn't get updated in my case. I had to update the IP manually. In the codepath you linked, IP always seems to equal lastIP in my case. Indeed my home's IP didn't change, what changed was the IP set on the provider. I'll add some debug logs now that is possible ^^ and send a report

Having an interface on top of logging third party is indeed a good thing. A few years ago I wrote a small tool in NodeJS that used Winston logger. I replaced it with Bunyan. I was proud because my tool had a noticeable speed increase. Then I got some user feedback complaining about Bunyan requiring Python. People who had not configured npm and proxies correctly wouldn't be able to download the binary, as a fallback npm was trying to compile the package and this would fail for these users... I had to rollback because I didn't write an interface and had no more time to spend on the tool..

qdm12:
I'm still working on the logging library with a few implementations, I should be done this weekend, and will update golibs in ddns-updater so that you can add debug logs where you please 😉

qdm12:
@fredericrous Have a look at https://github.com/qdm12/golibs/pull/3 about logging please 😉

Any feedback / suggestions / critics are welcome! I've been toying with it all my Saturday afternoon pretty much 😅

PR #170, opened 2021-02-15
Golibs logger update

qdm12:
Ran it and seems to work nicely (after [a few fixes!](https://github.com/qdm12/golibs/commit/c711ebd3e56ac62ec60c957865fe534cf4ca05a2))

Then feel free to do a PR to add debug logs where you would want some 😉  You can also directly branch off this repository if you want too, if that's easier for you.


fredericrous (commented):
I come after the battle but, approved, nice work. Maybe in the readme change the description  `Show caller per log line` by something more user friendly. What is a caller?
Regarding LOG_ENCODING. I think getting the logs as json was indeed not used by anyone, we'll see if someone opens an issue xD

Issue #91, opened 2020-08-06
Bug: Godaddy is bad 


oozj:
2020-08-06T15:21:03.055Z	ERROR	HTTP status 401 - Unauthorized : Could not authenticate API key/secret

qdm12:
Are you sure your API secret is valid?

jpherrell:
I'm having the same issue. I can access the API myself using Postman and the same keys. Doesn't fail if there is nothing to update, but does if there is a change.

qdm12:
Ok I'll check, maybe they changed their API. Should be done tomorrow.

qdm12:
Hello both, I checked and their API did not change wether to update DNS records or for authorization.

Unfortunately, I have no domain at GoDaddy anymore, can you please try

```sh
docker run -it --rm alpine:3.12
apk add curl
curl -X PUT -H "Authorization: sso-key [KEY]:[SECRET]" -H "Content-Type: application/json" -d '{"data":"127.0.01"}' "https://api.godaddy.com/v1/domains/[DOMAIN]/A/@"
exit
```

Replacing:

- `[KEY]` by your API Key
- `[SECRET]` by your API secret
- `[DOMAIN]` by your domain
- Eventually `@` by your host (or subdomain)

And see/report back the response? That should update your record to 127.0.0.1 (nothing really). That should work according to Godaddy's doc and ddns-updater *should* do the exact same.

jpherrell:
Sorry it has taken so long for me to reply:

When I send a body of {"data":"127.0.0.1"} I get this from them:
{
    "code": "INVALID_BODY",
    "fields": [
        {
            "code": "UNEXPECTED_TYPE",
            "message": "is not a array",
            "path": "records"
        }
    ],
    "message":
"Request body doesn't fulfill schema, see details in `fields`"
}

When I send a body of
[
    {
        "data": "127.0.0.1",
        "name": "@",
        "ttl": 3600,
        "type": "A"
    }
]

it works fine.

qdm12:
:+1: That should be it, will fix it tonight. Thanks for answering back!

qdm12:
Actually the program already sends `[{"data":"127.0.0.1"}]` as the body, can you try with just that body?
Referring to their API documentation, precisely to https://developer.godaddy.com/doc/endpoint/domains#/v1/recordReplaceTypeName, it seems only `"data"` is required in the body of the request; `"name'` and `"type"` are given in the URL.

[...]

PR #192, opened 2021-04-04
fix(config): Read case sensitive CONFIG

mchill:
Read the case sensitive value of the config file when using the CONFIG environment variable.
Converting the string to lowercase breaks case sensitive api keys/secrets for GoDaddy (and possibly
other providers).

Signed-off-by: Matthew Hill <matthewchill7@gmail.com>

Possibly solves #91.


qdm12 (approved):
You totally solved that, thanks! Plus good job on finding that case sensitive option 👍

PR #209, opened 2021-05-20
Feature: Aliyun DNS support

zzzhouuu:
add [aliyun](https://www.aliyun.com) provider


fredericrous:
It looks like there is an issue with the CI.
Also, I personally am in favor of using the sdk when possible. Using the package created by the vendor to interact with its API seems healthy to me. 1) It means less code to write (~ a third). 2) the day the API changes or the auth algorithm changes, I expect the lib to change this without me noticing. 3) If there is a breaking change and we don't want to continue with the lib, we can replace it, but in the midtime we get the feature.

qdm12:
> Also, I personally am in favor of using the sdk when possible. Using the package created by the vendor to interact with its API seems healthy to me. 1) It means less code to write (~ a third). 2) the day the API changes or the auth algorithm changes, I expect the lib to change this without me noticing. 3) If there is a breaking change and we don't want to continue with the lib, we can replace it, but in the midtime we get the feature.

I just got rid of the google gcp sdk dependency in a84a1d664a02b09b9313cb13e1bf8a1acbbb73b1 😄 My take on this:

1. It does add maybe 100 lines of code to write without an SDK, the biggest problem is the time spent to understand their SDK which is often hard to follow and full of options (it took me 4 hours for GCP!).
2. The API will likely stay retro-compatible, since not everyone will update their SDK dependency. So having your own code should still work the same too.
3. My biggest issues with SDKs are:
    1. added complexity, i.e. GCP SDK does read environment variables, without even asking it to
    2. added size, i.e. removing the GCP SDK reduced the image size from 17 to 11.6MB - not a big deal, but this is **one** SDK, so cumulating multiple SDKs might lead to a rather fat program with most of the code being unneeded (this is despite the Go compiler pruning code too!).

PR #215, opened 2021-06-16
Update README.md

opencsi-leon:
i try this solution for solve the permission write problem to volume


qdm12:
There is already this information above:

> You could change the user ID, for example with `1001`, by running the container with `--user=1001`)

But thanks for trying to improve the documentation nonetheless!

Issue #216, opened 2021-06-17
Feature request: add porkbun provider


fredericrous:
1. What's the feature?

Add porkbun.com as a provider

2. Extra information?

https://porkbun.com/api/json/v3/documentation

fredericrous:
I'll be working on this in the coming days, I'll do the alidns support first

qdm12:
Yes go for it. I have a branch locally for aliyun (Alibaba) but there isn't a hurry since he wrote the changes himself for it to work. I'll finish it without the dependency on their go api in the coming days. So feel free to dev porkbun (never heard of it, what a funny name!) if you want!

PR #217, opened 2021-07-18
add: porkbun provider

fredericrous:
resolves https://github.com/qdm12/ddns-updater/issues/216


fredericrous:
lol, I broke your github action @qdm12

qdm12:
> lol, I broke your github action @qdm12

😄  yes it's because of the branch name, docker tags don't accept `/`

As in it passed the verify stage so that's what matters mostly. I'll review your code now, thanks for the contribution!

internal/settings/providers/porkbun/provider.go
```diff
@@ -0,0 +1,295 @@
...
+	if err != nil {
+		return nil, err
+	}
+	p.setHeaders(request)
+
+	response, err := client.Do(request)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s", errors.ErrUnsuccessfulResponse, err)
+	}
+
```
qdm12, line 139: Put the `defer response.Body.Close()` here right after the error checking as it won't close it otherwise if the status code is not OK

internal/settings/providers/porkbun/provider.go
```diff
@@ -0,0 +1,295 @@
...
+			Name    string `json:"title"`
+			Type    string `json:"type"`
+			Content string `json:"content"`
+			TTL     string `json:"ttl"`
+			Prio    string `json:"prio"`
+			Notes   string `json:"notes"`
+		} `json:"records"`
+	}
+	defer response.Body.Close()
+	var responseData domainRecords
```
qdm12, line 158: - inline type
- remove `defer response.Body.Close()` in favor of moving it up
- Trim down to required fields for the record struct

```suggestion
	var responseData struct {
		Records []struct {
			Id      string `json:"id"`
			Content string `json:"content"`
		} `json:"records"`
	}
```

internal/settings/providers/porkbun/provider.go
```diff
@@ -0,0 +1,295 @@
...
+		return nil, fmt.Errorf("%w: %s", errors.ErrUnmarshalResponse, err)
+	}
+	var _recordIDs []string
+	for _, recordID := range responseData.Records {
+		if strings.HasSuffix(recordID.Content, p.domain) {
+			_recordIDs = append(_recordIDs, recordID.Id)
+		}
+	}
+	p.logger.Debug("getRecords: " + strings.Join(_recordIDs, ", "))
+	return _recordIDs, nil
```
qdm12, line 170: You can use `recordIDs` directly. Also adding some empty lines here and there.

Also `recordID` should be `record` I think

```suggestion

	for _, record := range responseData.Records {
		if strings.HasSuffix(record.Content, p.domain) {
			recordIDs = append(recordIDs, record.Id)
		}
	}

	p.logger.Debug("getRecords: " + strings.Join(recordIDs, ", "))
	return recordIDs, nil
```

internal/settings/providers/porkbun/provider.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+func (p *provider) createRecord(ctx context.Context, client *http.Client,
+	recordType string, ipStr string) (err error) {
+	u := url.URL{
+		Scheme: "https",
+		Host:   "porkbun.com",
+		Path:   fmt.Sprintf("/api/json/v3/dns/create/%s", p.domain),
```
qdm12, line 178:
```suggestion
		Path:   "/api/json/v3/dns/create/" + p.domain,
```

internal/settings/providers/porkbun/provider.go
```diff
@@ -0,0 +1,295 @@
...
+		TTL:          fmt.Sprint(p.ttl),
+	}
+	bodyBytes, err := json.Marshal(postRecordsParams)
+	if err != nil {
+		return fmt.Errorf("%w: %s", errors.ErrRequestMarshal, err)
+	}
+
+	p.logger.Debug("HTTP POST createRecord: " + u.String() + ": " + string(bodyBytes))
+
+	request, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), bytes.NewBuffer(bodyBytes))
```
qdm12, line 202: same thing about the json encoding directly to a bytes buffer 😉

internal/settings/providers/porkbun/provider.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	return nil
+}
+
+func (p *provider) updateRecord(ctx context.Context, client *http.Client,
+	recordType string, ipStr string, recordID string) (err error) {
+	u := url.URL{
+		Scheme: "https",
+		Host:   "porkbun.com",
+		Path:   fmt.Sprintf("/api/json/v3/dns/edit/%s/%s", p.domain, recordID),
```
qdm12, line 226:
```suggestion
		Path:   "/api/json/v3/dns/edit/" + p.domain + "/" + recordID),
```

internal/settings/providers/porkbun/provider.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+func (p *provider) Update(ctx context.Context, client *http.Client, ip net.IP) (newIP net.IP, err error) {
+	recordType := constants.A
+	var ipStr string
+	if ip.To4() == nil { // IPv6
+		recordType = constants.AAAA
+		ipStr = ip.To16().String()
+	} else {
+		ipStr = ip.To4().String()
```
qdm12, line 276: Do we need these conversions? Can we just do perhaps `ipStr := ip.String()` directly?

internal/settings/providers/porkbun/provider.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+}
+
+func (p *provider) setHeaders(request *http.Request) {
+	headers.SetUserAgent(request)
+	headers.SetContentType(request, "application/json")
+	headers.SetAccept(request, "application/json")
+}
+
+func (p *provider) getRecords(ctx context.Context, client *http.Client) (recordIDs []string, err error) {
```
qdm12, line 109: Rename this to `getRecordIDs`

internal/settings/providers/porkbun/provider.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+	} else {
+		for _, recordID := range recordIDs {
+			if err := p.updateRecord(ctx, client, recordType, ipStr, recordID); err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	return ip, nil
```
qdm12, line 294: cosmetic: reducing indentation

```suggestion
	if len(recordIDs) == 0 {
		if err := p.createRecord(ctx, client, recordType, ipStr); err != nil {
			return nil, err
		}
		return ip, nil
	}

	for _, recordID := range recordIDs {
		if err := p.updateRecord(ctx, client, recordType, ipStr, recordID); err != nil {
			return nil, err
		}
	}

	return ip, nil
```

qdm12 (commented):
Nice thanks for the contribution! A few comments, hopefully it's not too much 😉

docs/porkbun.md
```diff
@@ -0,0 +1,34 @@
...
+
+```json
+{
+  "settings": [
+    {
+      "provider": "porkbun",
+      "domain": "domain.com",
+      "host": "@",
+      "apikey": "sk1_7d119e3f656b00ae042980302e1425a04163c476efec1833q3cb0w54fc6f5022",
+      "secretapikey": "pk1_5299b57125c8f3cdf347d2fe0e713311ee3a1e11f11a14942b26472593e35368",
```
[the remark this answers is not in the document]
fredericrous: yes these are fake

fredericrous:
Thanks for the review, I'll continue to integrate these feedbacks this evening. 
Regarding the container version. Is this bach? https://github.com/qdm12/ddns-updater/blob/bc1a5b206fafa69503754a48bff358fcfb170a70/.github/workflows/build.yml#L77 we could do something like `echo ::set-output name=version::${TAG##*:}`

qdm12:
> Is this bach? we could do something like echo ::set-output name=version::${TAG##*/} on line 77 there:

Yes it's bash, but instead try to not use slashes in branches next time 😉

It brings a whole load of problems for git, go etc. I don't recall exactly what it was but it was terrible, [another example](https://coderwall.com/p/qkofma/a-caution-about-git-branch-names-with-s)

fredericrous:
I integrated the changes, I think we're good. thanks for having merged the `/` PR. I used to agree with you but changed my mind mainly because the `/` is supported in gitkraken (and I think in Git Extension but I never used this one and I'm on mac). Plus the `/` pattern is ok if we agree on putting one slash and no more

docs/porkbun.md
```diff
@@ -0,0 +1,35 @@
...
+}
+```
+
+### Parameters
+
+- `"domain"`
+- `"host"` is your host and can be a subdomain, `"*"` or `"@"`
+- `"apikey"`
+- `"secretapikey"`
+- `"ttl"` (optional)
```
qdm12, line 28:
```suggestion
- `"ttl"` optional integer value corresponding to a number of seconds
```

internal/settings/errors/validation.go
```diff
@@ -3,12 +3,14 @@ package errors
 import "errors"
 
 var (
+	ErrEmptyApiKey             = errors.New("empty aPI key")
```
qdm12, line 6:
```suggestion
	ErrEmptyApiKey             = errors.New("empty API key")
```

internal/settings/errors/validation.go
```diff
@@ -3,12 +3,14 @@ package errors
...
 	ErrEmptyAppKey             = errors.New("empty app key")
 	ErrEmptyConsumerKey        = errors.New("empty consumer key")
 	ErrEmptyEmail              = errors.New("empty email")
 	ErrEmptyKey                = errors.New("empty key")
 	ErrEmptyName               = errors.New("empty name")
 	ErrEmptyPassword           = errors.New("empty password")
+	ErrEmptyApiSecret          = errors.New("empty api secret")
```
qdm12, line 13:
```suggestion
	ErrEmptyApiSecret          = errors.New("empty API secret")
```

qdm12 (approved):
Nice looking good! Just two tiny suggestions

Also you may have to `git rebase master` your branch so that the readme has lfs (and also the CI gets fixed) 😉

qdm12 (approved):
Nice thanks for the changes! A few additional points

1. can you rebase or merge the readme from master so we can merge it please
2. If you don't mind, can you sneak in the `/` fix you had to the verify job for the publish job as well? If it's too complicated, then we can skip it.

Thanks!!

PR #220, opened 2021-07-19
Fix Readme line endings to lf

fredericrous:
Just converting those lines to lf


fredericrous:
I will just merge that because it's problematic with my git config. And I don't know how to solve this easily. Feel free to revert this commit

qdm12:
Thanks! Sorry my bad, I probably pushed it with \r\n by mistake a long time ago.

PR #249, opened 2021-10-01
Match configured host for LuaDNS instead of first A/AAAA record

lymanepp:
I don't have an easy way to test this, but the change was loosely lifted from the dreamhost provider


qdm12 (approved):
That should work 👍 

If you don't mind, before merging it, have a go using this command to build the image (you need `git` installed):

```sh
docker build -t qmcgaw/ddns-updater https://github.com/lymanepp/ddns-updater.git#luadns-host
```

It should build relatively quickly too. And then you can use it as you did, since it's also named `qmcgaw/ddns-updater`.

lymanepp:
I'm a newbie when it comes to Docker, but I was finally able to get it working. Good thing too, because there was a bug. All fixed and tested now.

PR #259, opened 2021-11-06
Add GHCR login and tag

adamus1red:
Publish containers to the Github Container Repo at the same time as docker hub

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,6 +63,13 @@ jobs:
           username: qmcgaw
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
 
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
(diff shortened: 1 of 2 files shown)


qdm12 (approved):
Awesome :100: !

Do you mind documenting in the Readme please? You can just add there is a mirror to the Docker Hub image in a one-liner, that should do :+1:

fredericrous:
just an honest question: why would you want to publish the image on 2 different spots? what advantage does it bring in your experience?

qdm12:
@fredericrous I guess as a backup, if Docker Hub goes down (it does, although rarely); Also Docker Hub has rate limiting (not a problem personally), but it can be problematic if you update your images automatically and have multiple updates e.g. the same day.

adamus1red:
@qdm12 that should be the README updated.

qdm12 (approved):

qdm12:
Awesome, thanks it's merged!!

PR #260, opened 2021-11-07
Add caching to improve build time performance

adamus1red:
Uses the builtin `cache-from`/`cache-to` options from the `docker/buil-push-action` to write out the buildx cache.
Using the github `actions/cache` action the buildx cache is stored by github for future runs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,8 +58,13 @@ jobs:
         with:
           target: base
           load: true
+          outputs: type=docker
+          build-args: |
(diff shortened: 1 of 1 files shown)


qdm12:
Hey there, thanks for your PR!

However, having toyed around with buildx caching a few months ago, it always turned out to be slower than without cache, at least for Go projects where things are usually quick to download/test/build. What do you think? I can merge it and compare publish time by doing another commit on top too.

If that turns out to be faster, I can modify the `verify` job to use buildx and we can try adding caching there too.

PR #270, opened 2021-11-25
Fix DDNSS provider specific IPv6 update flow

quantum-byte:
Fixed sending of specific ipv6 information to DDNSS provider. 

The DDNSS provider only has one keyword (`ip`) for specifying the new ip (see [documentation](https://ddnss.de/ua/help.php).

Currently when an ipv6 ip would be provided the `ip6` keyword is filled. This leads to the behaviour that DDNSS picks the ip to fill since the `ip` keyword is missing. This is not expected behaviour if `useProviderIP=false`.

I tested it with my own account/setup. Feel free to validate with something like the following:
`curl -i "https://www.ddnss.de/upd.php?key=<APIKEY>&host=<HOSTNAME>&ip=<IPV6>"`


qdm12:
Thanks!

Although checking https://ddnss.de/info.php and then pressing on "External IP" still shows the ip6 parameter. Can you perhaps confirm with them their documentation is wrong?

quantum-byte:
@qdm12 Oh wow i did not see that. I ll check what i can do :smile:

quantum-byte:
I sent them a description of the issue and request to clarify/fix the documentation via their contact form. I ll give feedback once they respond.

quantum-byte:
Just a quick update. I got a response but it was not really helpful.

But it triggered another test on my end and i found out that the `ip6` parameter works correctly if you enable the `dual-stack` setting on your hostname. Then it correctly updates the second ipv6 address. Though if you have it disabled it no longer works.

I sent them another message if the `ip6` parameter should work also if the `dual-stack` stetting is disabled. Lets see what i get back.

qdm12:
Hey there any update? Should we just merge it as it is?

qdm12:
cd37ab16463475cc4b450bc011d39f273ee8e90b should address this problem you have: it adds a boolean setting `"dual_stack": false`. If it is `false`, the `ip` parameter is always used. If it is `true` and the IP sent is IPv6, then `ip6` is used.

I tried force pushing to your fork but I got denied permission unfortunately, so I pushed it myself (mentioning you in the commit) to the master branch.

qdm12:
Oh actually the forced push worked, sorry I don't know why my terminal was erroring although it succeeded. Anyway, let's consider this fixed 👍 (even if ddnss.de fixes it on their end)

quantum-byte:
Hi @qdm12, sorry for the late reply and thanks for actually implementing a fix.

It works and complies to their specification, even though its not specified anywhere. From the responses i got, they will also not change it ...

qdm12:
Awesome thanks for letting me know. Also thanks for contributing, it's always appreciated! 👍

PR #337, opened 2022-06-06
add a provider for google cloud dns

### LINKED ISSUE #336: Feature request: Support for Google Cloud DNS
1. What's the feature?

It would be nice to have support for [Google Cloud DNS](https://cloud.google.com/dns)

orbatschow:
closes #336


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/qdm12/ddns-updater
 
-go 1.17
+go 1.18
```
qdm12, line 3: Please keep it to 1.17 for now, since golangci-lint doesn't have all its linters updated to work with go 1.18.

docs/gcp.md
```diff
@@ -0,0 +1,39 @@
...
+      "host": "@",
+      "ip_version": "ipv4"
+    }
+  ]
+}
+```
+### GOOGLE_APPLICATION_CREDENTIALS
+
+**IMPORTANT:** This provider relies on the GOOGLE_APPLICATION_CREDENTIALS environment variable to bet set.
+Make sure the variable is pointing to a valid file, which holds credentials for a Google Cloud service account.
```
qdm12, line 24: Let's just have it bind mounted to a fixed file path for now?
I don't see the point of having it configurable with an env variable, this adds complexity to the user.

internal/settings/errors/validation.go
```diff
@@ -13,11 +13,13 @@ var (
 	ErrEmptyApiSecret          = errors.New("empty API secret")
 	ErrEmptySecret             = errors.New("empty secret")
 	ErrEmptyToken              = errors.New("empty token")
+	ErrEmptyCredentials        = errors.New("environment variable GOOGLE_APPLICATION_CREDENTIALS not present")
 	ErrEmptyAccessKeyId        = errors.New("empty access key id")
 	ErrEmptyAccessKeySecret    = errors.New("empty key secret")
 	ErrEmptyTTL                = errors.New("TTL is not set")
 	ErrEmptyUsername           = errors.New("empty username")
 	ErrEmptyZoneIdentifier     = errors.New("empty zone identifier")
+	ErrEmptyProject            = errors.New("empty project")
```
qdm12, line 22: rename to `ErrGCPEmptyProject` since it's specific to GCP

internal/settings/providers/gcp/provider.go
```diff
@@ -0,0 +1,183 @@
...
+
+	dnsClient, err := clouddns.NewService(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	record, err := p.getRecord(dnsClient, recordType)
+	if err != nil && err != errors.ErrNotFound {
+		return nil, err
+	}
```
qdm12, line 112: I know this works fine, but I'm fighting hard to have the least dependency as possible.
Can these few calls be made without the sdk package?

Otherwise I'll look into it, feel free to let me know.

internal/settings/providers/gcp/provider.go
```diff
@@ -0,0 +1,183 @@
...
+		recordType = constants.AAAA
+	}
+
+	dnsClient, err := clouddns.NewService(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	record, err := p.getRecord(dnsClient, recordType)
+	if err != nil && err != errors.ErrNotFound {
```
qdm12, line 110: so that the error can be wrapped with context

```suggestion
	if err != nil && !errors.Is(err, errors.ErrNotFound) {
```

internal/settings/settings.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/qdm12/ddns-updater/internal/settings/providers/gcp"
```
qdm12, line 8: Please move this below sorted with the imports below

qdm12 (changes requested):
Thanks for the PR! Sorry for the delay, I had quite a backlog of things :wink: 

The main change would be to do all this without the GCP SDK.
Isn't there some documentation on how to interact with their API over http for example?

qdm12 (approved):
I rebased your branch and pushed changes. Notably:

- I kept the GCP SDK Go dependency as it was (bumped it up though)
- You need to place the content of your credentials json file in the config.json at the `"credentials"` key, as documented.

internal/settings/providers/gcp/provider.go
orbatschow: @qdm12 

I can see, that the old code `dnsClient, err := clouddns.NewService(ctx)`

changed to `clouddns.NewService(ctx, option.WithHTTPClient(client), option.WithCredentialsJSON(p.credentials))`

I can't tell who made this change, but i can't authenticate properly when `option.WithHTTPClient(client)` is active.

I also searched for the documentation regarding `WithHTTPClient` option, but could not manage to find anything describing the difference. The official docs state, this is the way to properly authenticate: https://github.com/googleapis/google-cloud-go#authorization

But they don't mention the HTTP client option.

When I remove the `WithHTTPClient` option authorization is fully working, otherwise I receive this error:

```
ERROR getting record resource set: googleapi: Error 401: Request is missing required authentication credential. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.
```

I am happy to provide a PR that just removes the usage of the client, what do you think?

qdm12: Oh my god!! I cannot believe I skipped your response. Thanks for clarifying!
Indeed, reading the google sdk implementation code, using the `WithHTTPClient` cancels all other options so the authentication option no longer works. However, I re-invented the wheel in a84a1d664a02b09b9313cb13e1bf8a1acbbb73b1 and now it *should* work with the custom client and authentication (oauth2) with GCP.

PR #353, opened 2022-08-23
Bumps versions of all GitHub actions

stumpylog:
This PR update all the actions to the latest current versions.  Naturally, the publish step fails for me, as I don't have secrets, etc setup for login.  It looks like 95% of the major version bumps are for using Node 16 as the default runtime, which isn't an issue for running hosted here.

The labeler action had the largest bump, so I updated it a tiny bit to account for the token being an input now (changed in [v3](https://github.com/crazy-max/ghaction-github-labeler/releases/tag/v3.0.0))

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
     env:
       DOCKER_BUILDKIT: "1"
     steps:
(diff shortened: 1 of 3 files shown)


qdm12 (approved):
Nice, thanks 👍

PR #354, opened 2022-08-23
Updates Alpine, golangci and xcputranslate versions

stumpylog:
Basically as it says in the title, updates versions of images this image depends on.

For golangci, I bumped it as far as possible, but [v1.48.0](https://github.com/golangci/golangci-lint/releases/tag/v1.48.0) produced a lint which would require me to know more about Go than I do, so I left it at [v1.46.2](https://github.com/golangci/golangci-lint/releases/tag/v1.46.2)


qdm12:
Tiny FYI this PR broke cross compilation (#359) due to bumping xcputranslate to v0.7.0 which has a different API than v0.6.0 😢 

Fixed it in f620e2a91cf940ab861bb34b62832734cf900a34

qdm12:
It also turns out my xcputranslate v0.7.0 is just plain wrong, I reverted back to v0.6.0 in 251112697a3ce577b0fe8ec215350ed737e48f77 so really shame on me 😅 !

PR #369, opened 2022-09-26
Improve UI

MrNavaStar:
Nothing crazy here, just made things feel a bit cleaner. Might try and get the favicon to work at some point as it currently does not.

### DIFF
--- a/internal/server/ui/index.html
+++ b/internal/server/ui/index.html
@@ -1,62 +1,93 @@
-<html>
+<!DOCTYPE html>
+<html lang="en">
 
 <head>
   <title>DDNS Updater</title>
   <link rel="icon" href="favicon.ico" type="image/x-icon">
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
   <style>
     table {
-      font-family: arial, sans-serif;
-      font-size: 14px;
-      font-size: 1vw;
       border-collapse: collapse;
+      background: white;
+      border-radius: 6px;
+      overflow: hidden;
+      max-width: 1200px;
       width: 100%;
+      margin: 0 auto;
+      position: center;
     }
 
-    td,
-    th {
-      border: 2px solid #9a9fa1;
+    table * {
+      position: relative;
+    }
+
+    table td,
+    table th {
+      padding-left: 8px;
+    }
+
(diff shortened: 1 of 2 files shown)


MrNavaStar:
Here's what it looks like, lmk what you think
image

internal/server/ui/index.html
```diff
@@ -1,62 +1,84 @@
-<html>
+<!DOCTYPE html>
+<html lang="en">
```
qdm12, line 2: nitpicking here; Are these necessary?

```suggestion
<html>

```

internal/server/ui/index.html
```diff
@@ -1,62 +1,84 @@
...
       <th>Host</th>
       <th>Provider</th>
-      <th>IP version</th>
-      <th>Update status</th>
-      <th>Set IP</th>
-      <th>Previous IPs (reverse chronological order)</th>
+      <th>IP Version</th>
+      <th>Update Status</th>
+      <th>Current IP</th>
+      <th>Previous IPs</th>
```
qdm12, line 81: Can we keep the `(reverse chronological order)` somehow? :thinking:

qdm12 (commented):
Thank you so much!
Also excuse my late review, I was a bit way-over-busy recently.

Did you check how it looks like on mobile? It would be nice to have something scaling on both desktop and mobile ideally :wink:

internal/server/ui/index.html
MrNavaStar: No doctype sometimes causes errors in the browser console, the lang probably isn't but it is good as we only support 1 localization

internal/server/ui/index.html
MrNavaStar: Yeah we can, it just looked cluttered and seems obvious in my opinion

MrNavaStar:
I didn't look on mobile, kinda forgot about that as I don't use my phone much for those sorts of things. A mobile ui will probably need to be quite different as viewing tables on mobile is always a pain

internal/server/ui/index.html
MrNavaStar: Maybe we could make it show on hover, that would probably be the best of both worlds

fuse314:
Consider not linking the whole font-awesome library just for the Github icon.
Instead, please add the SVG code of the logo directly to the HTML file.

PR #411, opened 2023-01-08
docs(readme): document `SHOUTRRR_PARAMS` environment variable

DennisGaida:
Found the environment variable SHOUTRRR_PARAMS in the code but not documented.

It is also necessary because by default ddns-updater sends plain text via Shoutrrr but sets the content type to `application/json` (invalid).

Even better would be a small usage example for Shoutrrr such as:

```
    --env SHOUTRRR_ADDRESSES="generic://<shoutrrr-webhook-endpoint-url>" \
    --env SHOUTRRR_PARAMS="contenttype=text/plain" \
```
but I didn't know where to add this to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,7 @@ Note that:
 | `LOG_LEVEL` | `info` | Level of logging, `debug`, `info`, `warning` or `error` |
 | `LOG_CALLER` | `hidden` | Show caller per log line, `hidden` or `short` |
 | `SHOUTRRR_ADDRESSES` |  | (optional) Comma separated list of [Shoutrrr addresses](https://containrrr.dev/shoutrrr/services/overview/) (notification services) |
+| `SHOUTRRR_PARAMS` |  | (optional) Comma separated list of Shoutrrr parameters (different per service). Set this to `contenttype=text/plain` when using the Generic Shoutrrr service since the default is `application/json`, but ddns-updater sends plain text.  |
 | `TZ` | | Timezone to have accurate times, i.e. `America/Montreal` |
 
 #### Public IP
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ Note that:
 | `LOG_LEVEL` | `info` | Level of logging, `debug`, `info`, `warning` or `error` |
 | `LOG_CALLER` | `hidden` | Show caller per log line, `hidden` or `short` |
 | `SHOUTRRR_ADDRESSES` |  | (optional) Comma separated list of [Shoutrrr addresses](https://containrrr.dev/shoutrrr/services/overview/) (notification services) |
-| `SHOUTRRR_PARAMS` |  | (optional) Comma separated list of Shoutrrr parameters (different per service). Set this to `contenttype=text/plain` when using the Generic Shoutrrr service since the default is `application/json`, but ddns-updater sends plain text.  |
+| `SHOUTRRR_PARAMS` | `title=DDNS Updater` | (optional) Comma separated list of Shoutrrr parameters (different per service). |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -220,6 +220,7 @@ Note that:
 | `LOG_LEVEL` | `info` | Level of logging, `debug`, `info`, `warning` or `error` |
 | `LOG_CALLER` | `hidden` | Show caller per log line, `hidden` or `short` |
 | `SHOUTRRR_ADDRESSES` |  | (optional) Comma separated list of [Shoutrrr addresses](https://containrrr.dev/shoutrrr/services/overview/) (notification services) |
+| `SHOUTRRR_PARAMS` |  | (optional) Comma separated list of Shoutrrr parameters (different per service). Set this to `contenttype=text/plain` when using the Generic Shoutrrr service since the default is `application/json`, but ddns-updater sends plain text.  |
```
qdm12, line 223: - Added the default `title=DDNS Updater`
- Removed the second part since this is most likely no longer true with the recent update of containerrr/shoutrrr v0.7.0 5fff4f4a564f0b59ba4ab4db16e370fe26f42a39

```suggestion
| `SHOUTRRR_PARAMS` | `title=DDNS Updater` | (optional) Comma separated list of Shoutrrr parameters (different per service). |
```

qdm12:
Awesome thanks!! Sorry for the huge delay, I was quite very busy the past few months, trying my best to manage multiple repositories 😄 

> It is also necessary because by default ddns-updater sends plain text via Shoutrrr but sets the content type to `application/json` (invalid).

I would guess this was due to the old version of the shoutrrr Go package v0.5.1 we were using. This got bumped to v0.7.0 in 5fff4f4a564f0b59ba4ab4db16e370fe26f42a39 can you try pulling the latest image (or v2.5.0) and see if this is still the case? Since the default of `application/json` is not from this repository code 🤔 Thanks!!

qdm12:
I digged in the shoutrrr library:

https://github.com/containrrr/shoutrrr/blob/048e61a9cff206d185bff54662ae7e0f5df49046/pkg/services/generic/generic_config.go#L15

The content type defaults to `application/json` for the generic endpoint, whatever the message is 😢 
I noticed you wrote `(different per service)` about the params, does that mean we could set the content type for the generic service only to plaintext??

qdm12:
I'm thinking of changing the content type to json for everything, since it's the default for the generic endpoint.
However, after some tests, it looks like changing it to application/json (without changing the message string send to shoutrrr) still works with Gotify, because its shoutrrr implementation handles the json encoding. Now with the generic endpoint, would your generic endpoint work with `application/json` as content-type? What should be the message sent to it?? Or does it simply refuses application/json? 🤔

README.md
```diff
@@ -220,6 +220,7 @@ Note that:
 | `LOG_LEVEL` | `info` | Level of logging, `debug`, `info`, `warning` or `error` |
 | `LOG_CALLER` | `hidden` | Show caller per log line, `hidden` or `short` |
 | `SHOUTRRR_ADDRESSES` |  | (optional) Comma separated list of [Shoutrrr addresses](https://containrrr.dev/shoutrrr/services/overview/) (notification services) |
+| `SHOUTRRR_PARAMS` | `title=DDNS Updater` | (optional) Comma separated list of Shoutrrr parameters (different per service). |
```
qdm12, line 223:
```suggestion
| `SHOUTRRR_PARAMS` | `title=DDNS Updater` | (optional) Comma separated list of Shoutrrr parameters. ⚠️ #416 |
```

qdm12 (approved):

Issue #319, opened 2022-03-13
Bug: Invalid signature during record creation on OVH with mode: api


P4uLT:
**Bad Request**: *ERROR cannot list records: bad HTTP status: 400 Bad Request: Invalid signature:*

1. Is this urgent: Yes
2. DNS provider(s) you use: OVH
3. Program version:

    

    `Running version latest built on 2022-02-26T13:23:18Z (commit b1a69e0)`

4. What are you using to run the container: docker-compose
5. Extra information (optional)

Logs:

```log
ddns-updater     | 2022/03/13 17:09:24 INFO Found single setting to update record
ddns-updater     | 2022/03/13 17:09:24 INFO Reading history from database: domain MY_DOMAIN host @
ddns-updater     | 2022/03/13 17:09:24 INFO healthcheck server: listening on 127.0.0.1:9999
ddns-updater     | 2022/03/13 17:09:24 INFO backup: disabled
ddns-updater     | 2022/03/13 17:09:24 INFO http server: listening on :8000
ddns-updater     | 2022/03/13 17:09:24 DEBUG configured to fetch IP: v4 or v6: false, v4: true, v6: false
ddns-updater     | 2022/03/13 17:09:25 DEBUG your public IP address are: v4 or v6: <nil>, v4: MY_PUBLIC_IP, v6: <nil>
ddns-updater     | 2022/03/13 17:09:30 WARN cannot DNS resolve MY_DOMAIN after 5 tries: lookup MY_DOMAIN on MY_PRIVATE_DNS:53: no such host
ddns-updater     | 2022/03/13 17:09:30 INFO IPv4 address of MY_DOMAIN is <nil> and your IPv4 address is MY_PUBLIC_IP
ddns-updater     | 2022/03/13 17:09:30 INFO Updating record [domain: MY_DOMAIN | host: @ | provider: OVH] to use MY_PUBLIC_IP
ddns-updater     | 2022/03/13 17:09:30 DEBUG GET https://eu.api.ovh.com/1.0/auth/time | headers: Accept: application/json;charset=utf-8; X-Ovh-Application: jbOL8kmskRVZq6Da
[...]

PR #414, opened 2023-01-10
#319 Fix OVH Invalid Signature

edaubert:
So this is the fix I propose for #319 

I am not confident with my Go skill so feel free to adapt it if required or ask me to do it if needed.


fredericrous:
I had to revert. you did some java coding in go, that does not work..

Issue #389, opened 2022-11-10
Domain resolution is resolving to internal IP causing constant record updates


cybersteel8:
TLDR: Hostname resolution should explicitly use an external DNS provider

1. Is this urgent: No
2. DNS provider(s) you use: No-IP
3. Program version: `Running version latest built on 2022-11-03T09:49:13Z (commit d8de9d2)`
4. What are you using to run the container: Unraid docker app
5. Extra information:

My server's default DNS resolver is my router, and in my router's configuration I have the domain used here overridden to return the internal IP. This applies only inside the local network, so the domain when accessed externally points to my external IP, as is governed by No-IP.
This means that DNS resolution of the domain name using the default DNS server, eg. 192.168.0.1 (the gateway) will never return the correct IP. To resolve the domain, any external DNS server should be used, like 1.1.1.1 or something else.
This is confusing ddns-updater and causing the domain to never match the external IP and it is always updating it, every cooldown.
If ddns-updater could be configured to specifically use an external DNS server, such as google or cloudflare or quad9 to resolve the domain name, this would return the correct response.

Logs:

```log
INFO IPv4 address of <domain> is <internal IP> and your IPv4 address is <external ip>
```

Configuration file:

```json
{
  "settings": [
    {
      "provider": "noip",
      "domain": "<redacted>",
      "host": "@",
      "username": "<redacted>",
      "password": "<redacted>",
      "ip_version": "ipv4",
      "provider_ip": true
    }
  ]
}

```

Host OS: Unraid 6.11.1

affeldt28:
I resolved this by setting a custom dns on the container.

For example with docker run:
`docker run ... --dns=1.1.1.1 ...`

or with docker-compose: 
```docker-compose
version: "3.7"
services:
  ddns-updater:
  ...
    dns:
      - 1.1.1.1
  ...
```

cybersteel8:
That's good advice, and I gave it a shot, but sadly it doesn't seem to work (or isn't respecting that environment variable)
Is it possible via an environment variable? I may need to wrangle Unraid to apply this flag not as `-e` but instead as literally `--dns` in the `docker run` command. This is basically what Unraid is doing:

```
docker run
  -d
  [...]
  -e 'dns'='1.1.1.1'
  [...]
```
...which isn't what you meant, is it?

GordonFreemanK:
Thanks a lot! That works perfectly for me too.
[...]

PR #417, opened 2023-01-18
docs(readme): add configuration notice split-dns #389

affeldt28:
Adds a configuration notice to the readme.

If you configured your internal dns to point to an internal ip, to also use your services within your network. Ddns-updater will check against this ip and triggers an update to your configured provider. Due to the fact that this record will never be your public ip, this update will be done every time it checks if your ip is up to date.
Adding a public dns to your container will prevent this behavior, because then ddns-updater will not check against your internal dns.


qdm12:
First, thank you for the PR 👍 
Second sorry for the massive delay!!
Third, in a7755823ba4b2873fe20ac76c31c869686db8b82 I added documentation in the Readme for the variable `RESOLVER_ADDRESS` - here since January but undocumented weirdly - and linked to your issue #389 and that should work such that only the domain name lookup mechanism uses the DNS specified instead of the whole container, which is a bit better.
By default it still uses the Docker network DNS server, but it can be changed for example to `1.1.1.1:53`.

PR #431, opened 2023-02-06
Fix 319

edaubert:
So this is the fix I propose for https://github.com/qdm12/ddns-updater/issues/319

It is a redo of https://github.com/qdm12/ddns-updater/pull/414 but with linter now

I also execute manually all actions from `verify` step.

Hope it will be enough


qdm12 (approved):
Awesome, thanks 👍

Issue #218, opened 2021-07-19
`name.com`


WinteWu:
https://www.name.com/api-docs

jdiekhoff:
@qdm12 I believe we can close this out now?

qdm12:
Indeed 👍

Support added by #474 which is part of the latest image and will be part of release image tag v2.6.0

PR #474, opened 2023-05-10
feat: add name.com provider

jdiekhoff:
Implementing feature request #218 to add support for name.com as a provider.


internal/settings/providers/namecom/getrecord.go
```diff
@@ -0,0 +1,67 @@
...
+		Scheme: "https",
+		Host:   "api.name.com",
+		Path:   fmt.Sprintf("/v4/domains/%s/records", p.domain),
+		User:   url.UserPassword(p.username, p.password),
+	}
+
+	// by default GET request will return 1000 records.
+	request, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
+	if err != nil {
+		return -1, fmt.Errorf("%w: %w", errors.ErrBadRequest, err)
```
qdm12, line 26: nit you can return `0`, anyway the caller should discard it if an error is returned. `-1` looks like a special number the caller may have to handle in case of errors 🤔

internal/settings/providers/namecom/getrecord.go
```diff
@@ -0,0 +1,67 @@
...
+		return -1, fmt.Errorf("%w: %w", errors.ErrUnmarshalResponse, err)
+	}
+
+	if len(data.Records) >= 1 {
+		for _, record := range data.Records {
+			if record.Host == p.host && record.Type == recordType {
+				return record.RecordID, nil
+			}
+		}
+	}
```
qdm12, line 64:
```suggestion
	for _, record := range data.Records {
		if record.Host == p.host && record.Type == recordType {
			return record.RecordID, nil
		}
	}
```

internal/settings/providers/namecom/provider.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	if stderrors.Is(err, errors.ErrRecordNotFound) {
+		err = p.createRecord(ctx, client, ip)
+		if err != nil {
+			return nil, fmt.Errorf("%w: %w", errors.ErrCreateRecord, err)
+		}
+	} else if err != nil {
+		return nil, fmt.Errorf("%w: %w", errors.ErrGetRecordID, err)
+	}
```
qdm12, line 120: nit it would be to flip this around and handle the error first and return earlier (i.e `return ip, nil`) so we have:
- less code blocks/indentation
- no need to check for `recordID > 0`
- Go standard programming approach is to handle errors first

```suggestion
if stderrors.Is(err, errors.ErrRecordNotFound) {
		err = p.createRecord(ctx, client, ip)
		if err != nil {
			return nil, fmt.Errorf("%w: %w", errors.ErrCreateRecord, err)
		}
		return ip, nil
	} else if err != nil {
		return nil, fmt.Errorf("%w: %w", errors.ErrGetRecordID, err)
	}

	err = p.updateRecord(ctx, client, recordID, ip)
	if err != nil {
		return nil, fmt.Errorf("%w: %w", errors.ErrUpdateRecord, err)
	}
```

docs/name.com.md
```diff
@@ -0,0 +1,45 @@
...
+- `"password"`
+
+### Optional parameters
+
+- `"ttl"` is the time this record can be cached for in seconds. Name.com allows a minimum TTL of 300, or 5 minutes. Name.com defaults to 300 if not provided.
+- `"ip_version"` can be `ipv4` (A records) or `ipv6` (AAAA records), defaults to `ipv4 or ipv6`
+
+## Domain setup
+
+<a href="https://www.name.com"><img src="../readme/name.svg" alt="drawing" width="25%"/></a>
```
qdm12, line 36: That's a nice touch 👍 Although out of curiosity, what's the size of it? I would guess a KB or so which is perfectly fine, just be careful it's not a megabyte that would bloat the repo quite a bit 😉

jdiekhoff: Size is 3KB, should be good.

jdiekhoff:
@qdm12 review suggestions applied and fixed a lint error.

docs/name.com.md
```diff
@@ -0,0 +1,45 @@
...
+### Compulsory parameters
+
+- `"domain"`
+- `"host"` is your host and can be a subdomain, `"@"` or `"*"` generally
+- `"username"`
+- `"password"`
+
+### Optional parameters
+
+- `"ttl"` is the time this record can be cached for in seconds. Name.com allows a minimum TTL of 300, or 5 minutes. Name.com defaults to 300 if not provided.
```
qdm12, line 31: We should add a check in the `isValid` method for this minimum

docs/name.com.md
qdm12, line 36: Can you move the svg file to the docs directory? Otherwise we could end up with a lot svgs at the root directory

docs/name.com.md
```diff
@@ -0,0 +1,45 @@
...
+- `"ip_version"` can be `ipv4` (A records) or `ipv6` (AAAA records), defaults to `ipv4 or ipv6`
+
+## Domain setup
```
qdm12, line 34: Although I appreciate the passion for documentation and to guide the user better 💯 I don't really see this section as necessary, I think the user is well aware he should have an account at name.com and have a domain purchased (or how to search for one).

Let's remove it unless you have a compelling reason to keep it?? 🤔

docs/name.com.md
```diff
@@ -0,0 +1,45 @@
...
+<a href="https://www.name.com"><img src="../readme/name.svg" alt="drawing" width="25%"/></a>
+
+1. Create a Name.com account and buy a domain name - *example.com* as an example
+1. Login to Name.com at [https://www.name.com/account/login](https://www.name.com/account/login)
+1. Search for domains to purchase [https://www.name.com/domain/search](https://www.name.com/domain/search)
+
+## API Credentials
+
+1. Get your API token at [https://www.name.com/account/settings/api](https://www.name.com/account/settings/api)
+1. Your username & password for the configuration will be your account username & token you generate from the API settings page.
```
qdm12, line 45: Given they call it a `token` on name.com, I think we should change the `password` json field to be `token`

internal/settings/providers/namecom/createrecord.go
```diff
@@ -0,0 +1,66 @@
...
+	response, err := client.Do(request)
+	if err != nil {
+		return fmt.Errorf("doing HTTP request: %w", err)
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusOK {
+		return fmt.Errorf("%w: %d: %s",
+			errors.ErrBadHTTPStatus, response.StatusCode,
+			utils.BodyToSingleLine(response.Body))
```
qdm12, line 62: Let's parse the error message

> Any non-200 response will return an ErrorMessage JSON response. ErrorMessages have a "message" that gives a general error message st ring, and an optional "details" key which contains a string with additional information about the error. 

Common API errors are listed at https://www.name.com/api-docs and you should match them to existing errors (from `internal/settings/errors`)

internal/settings/providers/namecom/createrecord.go
```diff
@@ -0,0 +1,66 @@
...
+
+	bodyBytes, err := json.Marshal(postRecordsParams)
+	if err != nil {
+		return fmt.Errorf("%w: %w", errors.ErrRequestMarshal, err)
+	}
+
+	request, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), bytes.NewBuffer(bodyBytes))
+	if err != nil {
+		return fmt.Errorf("%w: %w", errors.ErrBadRequest, err)
+	}
```
qdm12, line 51: > Any API calls requiring the use of the HTTP POST or PUT methods (for example, [Search](https://www.name.com/Domains#Search) or [UpdateEmailForwarding](https://www.name.com/EmailForwardings#UpdateEmailForwarding)) must set the Content-Type header to application/json.

Let's set that header on the request, and also set user agent headers.

For example other providers use a `setHeaders` method (or it can be a function):

```go
func (p *Provider) setHeaders(request *http.Request) {
	headers.SetContentType(request, "application/json")
	headers.SetAccept(request, "application/json")
	headers.SetUserAgent(request)
}
```

internal/settings/providers/namecom/getrecord.go
```diff
@@ -0,0 +1,65 @@
...
+		return 0, fmt.Errorf("%w: %w", errors.ErrUnmarshalResponse, err)
+	}
+
+	for _, record := range data.Records {
+		if record.Host == p.host && record.Type == recordType {
+			return record.RecordID, nil
+		}
+	}
+
+	return 0, fmt.Errorf("%w", errors.ErrRecordNotFound)
```
qdm12, line 64: nit add how many records we searched through

```suggestion
	return 0, fmt.Errorf("%w: in %d records", errors.ErrRecordNotFound, len(data.Records))
```

internal/settings/providers/namecom/updaterecord.go
```diff
@@ -0,0 +1,66 @@
...
+
+	bodyBytes, err := json.Marshal(postRecordsParams)
+	if err != nil {
+		return fmt.Errorf("%w: %w", errors.ErrRequestMarshal, err)
+	}
+
+	request, err := http.NewRequestWithContext(ctx, http.MethodPut, u.String(), bytes.NewBuffer(bodyBytes))
+	if err != nil {
+		return fmt.Errorf("%w: %w", errors.ErrBadRequest, err)
+	}
```
qdm12, line 51: we need request headers

qdm12 (commented):
Re-reviewed, there are a few changes/additions needed, sorry I didn't dig too much on the first review. Thanks! 👍

jdiekhoff:
> Re-reviewed, there are a few changes/additions needed, sorry I didn't dig too much on the first review. Thanks! 👍

Will look into these over the next couple of days, appreciate the thoroughness.

docs/name.com.md
qdm12: Never mind, silly me I didn't realize it was in the readme directory.

qdm12:
I rebased the branch and solved the PR feedback I left 😉 

You can get changes on your fork with `git fetch && git reset --hard origin/namedotcom`, and then can you try the image see if it still works fine?
It could also be interesting to try it with wrong credentials to see how it handles errors. Thanks!!

jdiekhoff:
@qdm12 had to make a slight change in the create logic. Previously if a record was created after not being found, the logic was setup so that the update method would be called. This resulted in an error as no recordID is included at this point so the update will fail. 

Create_Followed_By_Bad_Update

Also checked on using wrong credentials and a clean message is displayed:

Bad_Auth

qdm12:
Awesome, thanks! Merging ahead! 😉

qdm12 (approved):

PR #496, opened 2023-06-24
feat(provider): add suport for deSEC

### LINKED ISSUE #475: Feature request: support dedyn
1. What's the feature?
Support for Desec / Dedyn IPv4 und IPv6?

2. Extra information?
Protocol = dyndns2
Update-Server: update.dedyn.io

biochron:
Fixes #475

### DIFF
--- a/docs/desec.md
+++ b/docs/desec.md
@@ -35,7 +35,3 @@
 ### Web
 
 [desec.io/domains](https://desec.io/domains)
-
-### API
-
-[desec.readthedocs.io/en/latest/dns/domains](https://desec.readthedocs.io/en/latest/dns/domains.html)
--- a/docs/desec.md
+++ b/docs/desec.md
@@ -1,0 +1,38 @@
+# deSEC
+
+## Configuration
+
+### Example
+
+```json
+{
+  "settings": [
+    {
+      "provider": "desec",
+      "domain": "dedyn.io",
+      "host": "host",
+      "token": "token",
+      "ip_version": "ipv4",
+      "provider_ip": false
+    }
+  ]
+}
+```
+
+### Compulsory parameters
+
+- `"domain"`
+- `"host"`
+- `"token"` is your token that you can create [here](https://desec.io/tokens)
+
+### Optional parameters
+
+- `"ip_version"` can be `ipv4` (A records) or `ipv6` (AAAA records), defaults to `ipv4 or ipv6`
+- `"provider_ip"` can be set to `true` to let your DNS provider determine your IPv4 address (and/or IPv6 address) automatically when you send an update request, without sending the new IP address detected by the program in the request.
+
+## Domain setup
+
+### Web
+[https://desec.io/domains](https://desec.io/domains)
+### API
+[https://desec.readthedocs.io/en/latest/dns/domains.html](https://desec.readthedocs.io/en/latest/dns/domains.html)
--- a/docs/desec.md
+++ b/docs/desec.md
@@ -33,6 +33,9 @@
 ## Domain setup
 
 ### Web
-[https://desec.io/domains](https://desec.io/domains)
+
+[desec.io/domains](https://desec.io/domains)
+
 ### API
-[https://desec.readthedocs.io/en/latest/dns/domains.html](https://desec.readthedocs.io/en/latest/dns/domains.html)
+
+[desec.readthedocs.io/en/latest/dns/domains](https://desec.readthedocs.io/en/latest/dns/domains.html)
--- a/docs/desec.md
+++ b/docs/desec.md
@@ -32,6 +32,4 @@
 
 ## Domain setup
 
-### Web
-
 [desec.io/domains](https://desec.io/domains)
--- a/internal/provider/providers/desec/provider.go
+++ b/internal/provider/providers/desec/provider.go
@@ -125,9 +125,14 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
 	}
 	s := string(b)
 
-	if response.StatusCode != http.StatusOK {
-		return netip.Addr{}, fmt.Errorf("%w: %d: %s",
-			errors.ErrHTTPStatusNotValid, response.StatusCode, utils.ToSingleLine(s))
+	switch response.StatusCode {
+	case http.StatusOK:
+	case http.StatusUnauthorized:
+		return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrAuth, utils.ToSingleLine(s))
+	case http.StatusNotFound:
+		return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrHostnameNotExists, utils.ToSingleLine(s))
+	default:
+		return netip.Addr{}, fmt.Errorf("%w: %d: %s", errors.ErrHTTPStatusNotValid, response.StatusCode, utils.ToSingleLine(s))
 	}
 
 	switch {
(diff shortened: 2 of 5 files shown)


docs/desec.md
```diff
@@ -0,0 +1,38 @@
...
+
+- `"ip_version"` can be `ipv4` (A records) or `ipv6` (AAAA records), defaults to `ipv4 or ipv6`
+- `"provider_ip"` can be set to `true` to let your DNS provider determine your IPv4 address (and/or IPv6 address) automatically when you send an update request, without sending the new IP address detected by the program in the request.
+
+## Domain setup
+
+### Web
+[https://desec.io/domains](https://desec.io/domains)
+### API
+[https://desec.readthedocs.io/en/latest/dns/domains.html](https://desec.readthedocs.io/en/latest/dns/domains.html)
```
qdm12, line 38: nit add new lines to fit markdown formatting

```suggestion

[desec.io/domains](https://desec.io/domains)

### API

[desec.readthedocs.io/en/latest/dns/domains](https://desec.readthedocs.io/en/latest/dns/domains.html)
```

internal/provider/providers/desec/provider.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return p, nil
+}
+
+func (p *Provider) isValid() error {
+	switch {
+	case p.token == "":
+		return fmt.Errorf("%w", errors.ErrPasswordNotSet)
```
qdm12, line 56:
```suggestion
		return fmt.Errorf("%w", errors.ErrTokenNotSet)
```

internal/provider/providers/desec/provider.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	switch {
+	case strings.HasPrefix(s, constants.Notfqdn):
+		return netip.Addr{}, fmt.Errorf("%w", errors.ErrHostnameNotExists)
+	case strings.HasPrefix(s, "badrequest"):
+		return netip.Addr{}, fmt.Errorf("%w", errors.ErrBadRequest)
+	case strings.HasPrefix(s, "good"):
+		return ip, nil
+	default:
```
qdm12, line 140: Can you try with bad credentials and handle the error message here? 🙏

internal/provider/providers/desec/provider.go
```diff
@@ -0,0 +1,143 @@
...
+		return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrUnknownResponse, s)
```
qdm12, line 141:
```suggestion
		return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrUnknownResponse, utils.ToSingleLine(s))
```

qdm12 (commented):
Nice thanks 💯 
Just a few tiny comments to address before merging

internal/provider/providers/desec/provider.go
qdm12: Just have something like

```go
switch response.StatusCode {
case http.StatusOK:
case http.StatusUnauthorized:
	return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrAuth, utils.ToSingleLine(s))
default:
		errors.ErrHTTPStatusNotValid, response.StatusCode, utils.ToSingleLine(s))
}
```

to replace hat if block 😉  If the status is ok, it will execute the switch block below on the body content.

docs/desec.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Domain setup
+
+### Web
+
+[desec.io/domains](https://desec.io/domains)
+
+### API
+
+[desec.readthedocs.io/en/latest/dns/domains](https://desec.readthedocs.io/en/latest/dns/domains.html)
```
qdm12, line 41: I don't think this is needed to setup the domain from a user point of view right? If so, I would suggest removing it. You could move the link as a comment at the top of the Update method for the desec provider perhaps, although not obligatory either.

```suggestion
```

internal/provider/providers/desec/provider.go
```diff
@@ -0,0 +1,143 @@
...
+	if response.StatusCode != http.StatusOK {
+		return netip.Addr{}, fmt.Errorf("%w: %d: %s",
+			errors.ErrHTTPStatusNotValid, response.StatusCode, utils.ToSingleLine(s))
+	}
+
+	switch {
+	case strings.HasPrefix(s, constants.Notfqdn):
+		return netip.Addr{}, fmt.Errorf("%w", errors.ErrHostnameNotExists)
+	case strings.HasPrefix(s, "badrequest"):
+		return netip.Addr{}, fmt.Errorf("%w", errors.ErrBadRequest)
```
qdm12, line 137: Can you check if you send a malformed request, for example `myip=abc`, what's the status code returned (you can do that in your browser or using curl)? Is it 200 or could it be 400? Asking this since when unauthorized it returns the 401 status code... 🤔

biochron: I tried a malformed host ("hostname"=abc) and got a 404: nohost response. I also tried a malformed ip ("myip"=abc), and the server ignored my error, and correctly set the ip based on its own check ip function. I have added the new 404 to the status code check.
One last question before committing the patch. The string based error handler seems to be redundant, since the errors are handled through status codes. Should I remove the string based check from the commit?

qdm12: If their documentation mentions they use the dyndns protocol, I think we should keep it to match the dyndns protocol 🤔 Well technically we should support all codes as described in https://help.dyn.com/remote-access-api/return-codes/ but I am planning on refactoring all providers using the dyndns protocol and have a common single implementation for them, so don't bother I would say 😉

biochron: Sounds good, committed the patch with error handling.

internal/provider/providers/desec/provider.go
```diff
@@ -0,0 +1,148 @@
...
+	s := string(b)
+
+	switch response.StatusCode {
+	case http.StatusOK:
+	case http.StatusUnauthorized:
+		return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrAuth, utils.ToSingleLine(s))
+	case http.StatusNotFound:
+		return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrHostnameNotExists, utils.ToSingleLine(s))
+	default:
+		return netip.Addr{}, fmt.Errorf("%w: %d: %s", errors.ErrHTTPStatusNotValid, response.StatusCode, utils.ToSingleLine(s))
```
qdm12, line 135:
```suggestion
		return netip.Addr{}, fmt.Errorf("%w: %d: %s", errors.ErrHTTPStatusNotValid,
			response.StatusCode, utils.ToSingleLine(s))
```

docs/desec.md
```diff
@@ -0,0 +1,37 @@
...
+### Optional parameters
+
+- `"ip_version"` can be `ipv4` (A records) or `ipv6` (AAAA records), defaults to `ipv4 or ipv6`
+- `"provider_ip"` can be set to `true` to let your DNS provider determine your IPv4 address (and/or IPv6 address) automatically when you send an update request, without sending the new IP address detected by the program in the request.
+
+## Domain setup
+
+### Web
+
+[desec.io/domains](https://desec.io/domains)
```
qdm12, line 37:
```suggestion
[desec.io/domains](https://desec.io/domains)
```

qdm12 (approved):

qdm12:
Merged, thanks 🎖️

Issue #493, opened 2023-06-17
Feature request: Hetzner API


AbsusRex:
1. What's the feature?

Would love to see support for Hetzner

2. Extra information?
API doc: https://dns.hetzner.com/api-docs
And there's also this existing project https://github.com/kutzilla/docker-hetzner-ddns/tree/master

qdm12:
Done in 301e4d6cc4a4fe19da28a6c64025c6c840979ba8 for the latest Docker image. Thank you for your patience!!

PR #503, opened 2023-07-14
feat(provider): add suport for Hetzner

lieblinger:
Based on cloudflare provider and related to https://github.com/qdm12/ddns-updater/issues/493

### DIFF
--- a/internal/provider/providers/hetzner/create.go
+++ b/internal/provider/providers/hetzner/create.go
@@ -77,7 +77,7 @@ func (p *Provider) createRecord(ctx context.Context, client *http.Client, ip net
 	}
 
 	if parsedJSON.Record.ID == "" {
-		return fmt.Errorf("%w: %s", errors.ErrUnsuccessful, "error: empty response without id")
+		return fmt.Errorf("%w", errors.ErrReceivedNoResult)
 	}
 
 	return nil
--- a/internal/provider/providers/hetzner/create.go
+++ b/internal/provider/providers/hetzner/create.go
@@ -69,12 +69,17 @@ func (p *Provider) createRecord(ctx context.Context, client *http.Client, ip net
 	decoder := json.NewDecoder(response.Body)
 	var parsedJSON struct {
 		Record struct {
-			ID string `json:"id"`
+			ID    string     `json:"id"`
+			Value netip.Addr `json:"value"`
 		} `json:"record"`
 	}
 	err = decoder.Decode(&parsedJSON)
+	newIP := parsedJSON.Record.Value
 	if err != nil {
 		return fmt.Errorf("json decoding response body: %w", err)
+	} else if newIP.Compare(ip) != 0 {
+		return fmt.Errorf("%w: sent ip %s to update but received %s",
+			errors.ErrIPReceivedMismatch, ip, newIP)
 	}
 
 	if parsedJSON.Record.ID == "" {
--- a/internal/provider/providers/hetzner/create.go
+++ b/internal/provider/providers/hetzner/create.go
@@ -60,9 +60,10 @@ func (p *Provider) createRecord(ctx context.Context, client *http.Client, ip net
 	}
 	defer response.Body.Close()
 
-	if response.StatusCode > http.StatusUnsupportedMediaType {
+	if response.StatusCode != http.StatusOK {
 		return fmt.Errorf("%w: %d: %s",
-			errors.ErrHTTPStatusNotValid, response.StatusCode, utils.BodyToSingleLine(response.Body))
+			errors.ErrHTTPStatusNotValid, response.StatusCode,
+			utils.BodyToSingleLine(response.Body))
 	}
 
 	decoder := json.NewDecoder(response.Body)
--- a/internal/provider/providers/hetzner/create.go
+++ b/internal/provider/providers/hetzner/create.go
@@ -1,0 +1,84 @@
+package hetzner
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/netip"
+	"net/url"
+
+	"github.com/qdm12/ddns-updater/internal/provider/constants"
+	"github.com/qdm12/ddns-updater/internal/provider/errors"
+	"github.com/qdm12/ddns-updater/internal/provider/utils"
+)
+
+func (p *Provider) createRecord(ctx context.Context, client *http.Client, ip netip.Addr) (err error) {
+	recordType := constants.A
+	if ip.Is6() {
+		recordType = constants.AAAA
+	}
+
+	u := url.URL{
+		Scheme: "https",
+		Host:   "dns.hetzner.com",
+		Path:   fmt.Sprintf("/api/v1/records"),
+	}
+
+	requestData := struct {
+		Type           string `json:"type"`
+		Name           string `json:"name"`
+		Value          string `json:"value"`
+		ZoneIdentifier string `json:"zone_id"`
+		TTL            uint   `json:"ttl"`
+	}{
+		Type:           recordType,
+		Name:           p.host,
+		Value:          ip.String(),
+		ZoneIdentifier: p.zoneIdentifier,
+		TTL:            p.ttl,
+	}
+
+	buffer := bytes.NewBuffer(nil)
+	encoder := json.NewEncoder(buffer)
+	err = encoder.Encode(requestData)
+	if err != nil {
+		return fmt.Errorf("JSON encoding request data: %w", err)
+	}
+
+	request, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), buffer)
+	if err != nil {
+		return fmt.Errorf("creating http request: %w", err)
+	}
+
+	p.setHeaders(request)
+
+	response, err := client.Do(request)
+	if err != nil {
+		return err
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode > http.StatusUnsupportedMediaType {
+		return fmt.Errorf("%w: %d: %s",
+			errors.ErrHTTPStatusNotValid, response.StatusCode, utils.BodyToSingleLine(response.Body))
+	}
+
+	decoder := json.NewDecoder(response.Body)
+	var parsedJSON struct {
+		Record struct {
+			ID string `json:"id"`
+		} `json:"record"`
+	}
+	err = decoder.Decode(&parsedJSON)
+	if err != nil {
+		return fmt.Errorf("json decoding response body: %w", err)
+	}
+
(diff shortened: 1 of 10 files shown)


internal/provider/providers/hetzner/provider.go
```diff
@@ -0,0 +1,340 @@
...
+	u := url.URL{
+		Scheme: "https",
+		Host:   "dns.hetzner.com",
+		Path:   fmt.Sprintf("/api/v1/records"),
+	}
+
+	requestData := struct {
+		Type           string `json:"type"`  // constants.A or constants.AAAA depending on ip address given
+		Name           string `json:"name"`  // DNS record name i.e. example.com
+		Value          string `json:"value"` // ip address
```
qdm12, line 206: nit comments are not really needed, since we can refer to the struct fields initialization right below

```suggestion
		Type           string `json:"type"` 
		Name           string `json:"name"` 
		Value          string `json:"value"`
```

internal/provider/providers/hetzner/provider.go
```diff
@@ -0,0 +1,340 @@
...
+	case len(listRecordsResponse.Records) > 1:
+		return "", false, fmt.Errorf("%w: %d instead of 1",
+			errors.ErrResultsCountReceived, len(listRecordsResponse.Records))
+	case listRecordsResponse.Records[0].Value == newIP.String(): // up to date
+		return "", true, nil
+	}
+	return listRecordsResponse.Records[0].ID, false, nil
+}
+
+func (p *Provider) CreateRecord(ctx context.Context, client *http.Client, ip netip.Addr) (recordID string, err error) {
```
qdm12, line 190: this can be unexported, rename to `createRecord`

internal/provider/providers/hetzner/provider.go
```diff
@@ -0,0 +1,340 @@
...
+		Record struct {
+			ID string `json:"id"`
+		} `json:"record"`
+	}
+	err = decoder.Decode(&parsedJSON)
+	if err != nil {
+		return "", fmt.Errorf("json decoding response body: %w", err)
+	}
+
+	return parsedJSON.Record.ID, nil
```
qdm12, line 253: add a check to see if the record id is empty or not (i.e. if the server responds with `{}` the json decoding will work but the id will be left empty)

internal/provider/providers/hetzner/provider.go
```diff
@@ -0,0 +1,340 @@
...
+	u := url.URL{
+		Scheme: "https",
+		Host:   "dns.hetzner.com",
+		Path:   fmt.Sprintf("/api/v1/records/%s", identifier),
+	}
+
+	requestData := struct {
+		Type           string `json:"type"`  // constants.A or constants.AAAA depending on ip address given
+		Name           string `json:"name"`  // DNS record name i.e. example.com
+		Value          string `json:"value"` // ip address
```
qdm12, line 285:
```suggestion
		Type           string `json:"type"`
		Name           string `json:"name"`
		Value          string `json:"value"`
```

internal/provider/providers/hetzner/provider.go
```diff
@@ -0,0 +1,326 @@
...
+func (p *Provider) setHeaders(request *http.Request) {
+	headers.SetUserAgent(request)
+	headers.SetContentType(request, "application/json")
+	headers.SetAccept(request, "application/json")
+	if p.token != "" {
+		request.Header.Set("Auth-API-Token", p.token)
+	}
+}
+
+// Obtain domain ID.
```
qdm12, line 109: nit unneeded comment, the function name `getRecordID` explains it all 😉 

```suggestion
```

internal/provider/providers/hetzner/provider.go
```diff
@@ -0,0 +1,326 @@
...
+// See https://dns.hetzner.com/api-docs#operation/GetZones.
+func (p *Provider) getRecordID(ctx context.Context, client *http.Client, newIP netip.Addr) (
```
qdm12, line 111: Can you move this function to `getrecord.go`

internal/provider/providers/hetzner/provider.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+
+	if !parsedJSON.Record.ID {
+		return "", fmt.Errorf("%w: %s", errors.ErrUnsuccessful, "error: empty response without id")
+	}
+
+	return parsedJSON.Record.ID, nil
+}
+
+func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Addr) (newIP netip.Addr, err error) {
```
qdm12, line 242: nit can you move this function above the unexported functions `getRecordID` and `createRecord`

internal/provider/providers/hetzner/provider.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+
+	identifier, upToDate, err := p.getRecordID(ctx, client, ip)
+
+	switch {
+	case stderrors.Is(err, errors.ErrReceivedNoResult):
+		identifier, err = p.createRecord(ctx, client, ip)
+		if err != nil {
+			return netip.Addr{}, fmt.Errorf("creating record: %w", err)
+		}
```
qdm12, line 255: I think we can return if we create the record right?

```suggestion
		identifier, err = p.createRecord(ctx, client, ip)
		if err != nil {
			return netip.Addr{}, fmt.Errorf("creating record: %w", err)
		}
		return ip, nil
```

internal/provider/providers/hetzner/provider.go
```diff
@@ -0,0 +1,326 @@
...
+	case err != nil:
+		return netip.Addr{}, fmt.Errorf("getting record id: %w", err)
+	case upToDate:
+		return ip, nil
+	}
+
+	u := url.URL{
+		Scheme: "https",
+		Host:   "dns.hetzner.com",
```
qdm12, line 264: nit we could have this and the function body below in its own unexported function `updateRecord` to keep this tidy?

qdm12 (commented):
Thanks for the changes 🎖️! A few minor comments and it should be good to merge 😉

qdm12:
Also:

- linting is failing, try running `golangci-lint run` locally to check why? (or check the verify job log output)
- maybe, you could move each long function to its own file (createrecord.go, getrecord.go, updaterecord.go). Not 100% compulsory, but I used such split in other providers and it's nicer to navigate in my opinion.

lieblinger:
Thanks for your review! I've made some changes and hope that it's okay now. If there's anything else that needs to be change just let me know.

lieblinger:
@qdm12 Do we need to change anything else?

varac:
I tested this PR using an image built from this PR banch (`registry.0xacab.org/varac-projects/images/ddns-updater:hetzner`) and it works well :popcorn: 
Looking forward to get this merged soon.

internal/provider/providers/hetzner/update.go
```diff
@@ -0,0 +1,88 @@
...
+	response, err := client.Do(request)
+	if err != nil {
+		return netip.Addr{}, err
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode > http.StatusUnsupportedMediaType {
+		return netip.Addr{}, fmt.Errorf("%w: %d: %s",
+			errors.ErrHTTPStatusNotValid, response.StatusCode, utils.BodyToSingleLine(response.Body))
+	}
```
qdm12, line 67: same here, I think we should error on any 4xx codes 🤔

internal/provider/providers/hetzner/create.go
```diff
@@ -0,0 +1,84 @@
...
+	if response.StatusCode > http.StatusUnsupportedMediaType {
+		return fmt.Errorf("%w: %d: %s",
+			errors.ErrHTTPStatusNotValid, response.StatusCode, utils.BodyToSingleLine(response.Body))
+	}
+
+	decoder := json.NewDecoder(response.Body)
+	var parsedJSON struct {
+		Record struct {
+			ID string `json:"id"`
+		} `json:"record"`
```
qdm12, line 72: Can we also get the `"value"` field so that we can return the `newIP` from `createRecord`, as returned from the server? So that we can compare it with the ip we sent. A bit of a extreme cautionary measure, but better safe than sorry 😉

internal/provider/providers/hetzner/update.go
```diff
@@ -0,0 +1,88 @@
...
+
+	decoder := json.NewDecoder(response.Body)
+	var parsedJSON struct {
+		Record struct {
+			Value string `json:"value"`
```
qdm12, line 72: nit You can have 

```go
Value string `json:"value"`
```

directly as

```go
Value netip.Addr `json:"value"`
```

So that you can remove the netip parsing below (but keep the ip vs newip comparison)😉

qdm12 (commented):
Sorry for the massive delay reviewing 😢 
A final review, it shouldn't be too complex to fix, let me know your thoughts.
Feel free to only comment back on new conversations, I can finish up the coding fixes if you don't have time, especially since it has been a few weeks since this was opened sorry!

lieblinger:
Thanks for your review. All changes are made and I hope that everything is okay. 
If there's something else I can update just let me know.

internal/provider/providers/hetzner/getrecord.go
```diff
@@ -0,0 +1,77 @@
...
+		} `json:"records"`
+	}{}
+	err = decoder.Decode(&listRecordsResponse)
+	if err != nil {
+		return "", false, fmt.Errorf("json decoding response body: %w", err)
+	}
+
+	switch {
+	case len(listRecordsResponse.Records) == 0:
+		return "", false, fmt.Errorf("%w", errors.ErrReceivedNoResult)
```
qdm12, line 69: Quick double check question, if there is no record (we need to create it), is the http status code still ok ? Or could it be not found?

internal/provider/providers/hetzner/create.go
```diff
@@ -0,0 +1,90 @@
...
+		return fmt.Errorf("%w: %d: %s",
+			errors.ErrHTTPStatusNotValid, response.StatusCode,
+			utils.BodyToSingleLine(response.Body))
+	}
+
+	decoder := json.NewDecoder(response.Body)
+	var parsedJSON struct {
+		Record struct {
+			ID    string     `json:"id"`
+			Value netip.Addr `json:"value"`
```
qdm12, line 73: Are we sure the server responds with a `"value"` field in the create record call? I mentioned but didn't check 🥲

qdm12 (approved):
Awesome 💯 
Just two last tiny double checking questions 😉 And then let's merge finally 👍

varac:
@qdm12 Do you mind doing a new release with this feature ? Thanks!

PR #504, opened 2023-07-15
Added Now-DNS issue#368

### LINKED ISSUE #368: Feature request: support now-dns.com
1. What's the feature?
Add provider [now-dns](https://now-dns.com/)

2. Extra information?

[API Documentation](https://now-dns.com/?p=clients)

Here is an example of a Python script I made to automatically update DNS:

```python
import requests
import time
import creds

domain = creds.domain
username = creds.username
password = creds.password


def pushover_message(title, message):

    url = "https://api.pushover.net/1/messages.json"

    payload={'token': creds.app_id,
    'user': creds.user_id,
    'message': message,
    'title': title}

    requests.request("POST", url, data=payload)




while True:
    try:
        r = requests.get(f'https://now-dns.com/update?hostname={domain}', auth=(username, password))

        if r.text != "nochg":
            pushover_message("NOW-DDNS Alert", r.text)

        if r.status_code != 200:
            pushover_message("NOW-DDNS Alert", r.status_code)


        time.sleep(300)


    except Exception as e:
        print(e)
        pushover_message("NOW-DDNS Alert - Python Exception", e)
        time.sleep(300)
```

CyberAustin:
Fixes #368

### DIFF
--- a/docs/nowdns.md
+++ b/docs/nowdns.md
@@ -10,7 +10,6 @@
     {
       "provider": "nowdns",
       "domain": "domain.com",
-      "host": "@",
       "username": "username",
       "password": "password",
     }
@@ -20,7 +19,6 @@
 
 ### Compulsory parameters
 
-- `"domain"` your domain name
-- `"host"` leave it set to `"@"`
+- `"domain"` your full domain name (FQDN)
 - `"username"` your email address
 - `"password"`
--- a/docs/nowdns.md
+++ b/docs/nowdns.md
@@ -1,0 +1,26 @@
+# Now-DNS
+
+## Configuration
+
+### Example
+
+```json
+{
+  "settings": [
+    {
+      "provider": "nowdns",
+      "domain": "domain.com",
+      "host": "@",
+      "username": "username",
+      "password": "password",
+    }
+  ]
+}
+```
+
+### Compulsory parameters
+
+- `"domain"` your domain name
+- `"host"` leave it set to `"@"`
+- `"username"` your email address
+- `"password"`
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,9 @@ module github.com/qdm12/ddns-updater
 
 go 1.20
 
+//required for local testing. Linter will not build the docker container with it in place though. 
+//replace github.com/qdm12/ddns-updater/internal/provider/providers/nowdns => ./internal/provider/providers/nowdns
+
 require (
 	github.com/breml/rootcerts v0.2.11
 	github.com/containrrr/shoutrrr v0.7.0
--- a/go.mod
+++ b/go.mod
@@ -2,9 +2,6 @@ module github.com/qdm12/ddns-updater
 
 go 1.20
 
-//required for local testing. Linter will not build the docker container with it in place though. 
-//replace github.com/qdm12/ddns-updater/internal/provider/providers/nowdns => ./internal/provider/providers/nowdns
-
 require (
 	github.com/breml/rootcerts v0.2.11
 	github.com/containrrr/shoutrrr v0.7.0
--- a/internal/provider/constants/providers.go
+++ b/internal/provider/constants/providers.go
@@ -35,6 +35,7 @@ const (
 	Netcup       models.Provider = "netcup"
 	Njalla       models.Provider = "njalla"
 	NoIP         models.Provider = "noip"
+	NowDNS       models.Provider = "nowdns"
 	OpenDNS      models.Provider = "opendns"
 	OVH          models.Provider = "ovh"
 	Porkbun      models.Provider = "porkbun"
--- a/internal/provider/constants/providers.go
+++ b/internal/provider/constants/providers.go
@@ -79,6 +79,7 @@ func ProviderChoices() []models.Provider {
 		NameCom,
 		Njalla,
 		NoIP,
+		NowDNS,
 		OpenDNS,
 		OVH,
 		Porkbun,
--- a/internal/provider/providers/nowdns/provider.go
+++ b/internal/provider/providers/nowdns/provider.go
@@ -134,7 +134,6 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
 	case http.StatusOK:
 		switch {
 		case strings.Contains(s, "good"):
-			fmt.Println("GOOD!")
 			newIP, err = netip.ParseAddr(ip.String())
 			if err != nil {
 				return netip.Addr{}, fmt.Errorf("%w: %w", errors.ErrIPReceivedMalformed, err)
@@ -142,9 +141,8 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
 				return netip.Addr{}, fmt.Errorf("%w: sent ip %s to update but received %s",
 					errors.ErrIPReceivedMismatch, ip, newIP)
 			}
-			return newIP, nil
+			return ip, nil
 		case strings.Contains(s, "nochg"):
-			fmt.Println(s)
 			newIP, err = netip.ParseAddr(ip.String())
 			if err != nil {
 				return netip.Addr{}, fmt.Errorf("%w: in response %q", errors.ErrReceivedNoResult, s)
@@ -152,7 +150,7 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
 				return netip.Addr{}, fmt.Errorf("%w: sent ip %s to update but received %s",
 					errors.ErrIPReceivedMismatch, ip, newIP)
 			}
-			return newIP, nil
+			return ip, nil
 		default:
 			return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrUnknownResponse, s)
 		}
@@ -163,7 +161,7 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
 		case constants.Badauth:
 			return netip.Addr{}, fmt.Errorf("%w", errors.ErrAuth)
 		default:
-			return netip.Addr{}, fmt.Errorf("%w: %d: %s", errors.ErrHTTPStatusNotValid, response.StatusCode, s)
+			return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrUnknownResponse, s)
 		}
 	default:
 		return netip.Addr{}, fmt.Errorf("%w: %d: %s", errors.ErrHTTPStatusNotValid, response.StatusCode, s)
--- a/internal/provider/providers/nowdns/provider.go
+++ b/internal/provider/providers/nowdns/provider.go
@@ -59,8 +59,6 @@ func (p *Provider) isValid() error {
 		return fmt.Errorf("%w", errors.ErrUsernameNotSet)
 	case p.password == "":
 		return fmt.Errorf("%w", errors.ErrPasswordNotSet)
-	case p.host == "*":
-		return fmt.Errorf("%w", errors.ErrHostWildcard)
 	}
 	return nil
 }
--- a/internal/provider/providers/nowdns/provider.go
+++ b/internal/provider/providers/nowdns/provider.go
@@ -1,0 +1,169 @@
+package nowdns
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/netip"
+	"net/url"
+	"strings"
+
+	"github.com/qdm12/ddns-updater/internal/models"
+	"github.com/qdm12/ddns-updater/internal/provider/constants"
+	"github.com/qdm12/ddns-updater/internal/provider/errors"
+	"github.com/qdm12/ddns-updater/internal/provider/headers"
+	"github.com/qdm12/ddns-updater/internal/provider/utils"
+	"github.com/qdm12/ddns-updater/pkg/publicip/ipversion"
+)
+
+type Provider struct {
+	domain        string
+	host          string
+	ipVersion     ipversion.IPVersion
+	username      string
+	password      string
+	useProviderIP bool
+}
+
+func New(data json.RawMessage, domain, host string,
+	ipVersion ipversion.IPVersion) (p *Provider, err error) {
+	extraSettings := struct {
+		Username      string `json:"username"`
+		Password      string `json:"password"`
+		UseProviderIP bool   `json:"provider_ip"`
+	}{}
+	err = json.Unmarshal(data, &extraSettings)
+	if err != nil {
+		return nil, err
+	}
+	p = &Provider{
+		domain:        domain,
+		host:          host,
+		ipVersion:     ipVersion,
+		username:      extraSettings.Username,
+		password:      extraSettings.Password,
+		useProviderIP: extraSettings.UseProviderIP,
+	}
+	err = p.isValid()
+	if err != nil {
+		return nil, err
+	}
+	return p, nil
+}
+
+func (p *Provider) isValid() error {
+	switch {
+	case p.username == "":
+		return fmt.Errorf("%w", errors.ErrUsernameNotSet)
+	case p.password == "":
+		return fmt.Errorf("%w", errors.ErrPasswordNotSet)
+	case p.host == "*":
+		return fmt.Errorf("%w", errors.ErrHostWildcard)
+	}
+	return nil
+}
+
+func (p *Provider) String() string {
+	return utils.ToString(p.domain, p.host, constants.Infomaniak, p.ipVersion)
+}
+
+func (p *Provider) Domain() string {
+	return p.domain
+}
+
+func (p *Provider) Host() string {
+	return p.host
+}
+
+func (p *Provider) IPVersion() ipversion.IPVersion {
+	return p.ipVersion
+}
+
+func (p *Provider) Proxied() bool {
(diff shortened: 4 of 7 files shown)


CyberAustin:
Crap. Line 137 I left a "debugging statement" in there 😄

go.mod
```diff
@@ -2,6 +2,9 @@ module github.com/qdm12/ddns-updater
 
 go 1.20
 
+//required for local testing. Linter will not build the docker container with it in place though. 
+//replace github.com/qdm12/ddns-updater/internal/provider/providers/nowdns => ./internal/provider/providers/nowdns
+
```
qdm12, line 7: I don't think this is needed even for local testing, since the Go module is `github.com/qdm12/ddns-updater` (even your local version), so it will use your local files (also the `internal` directory cannot be imported from outside)

```suggestion
```

internal/provider/providers/nowdns/provider.go
```diff
@@ -0,0 +1,171 @@
...
+			newIP, err = netip.ParseAddr(ip.String())
+			if err != nil {
+				return netip.Addr{}, fmt.Errorf("%w: %w", errors.ErrIPReceivedMalformed, err)
+			} else if !p.useProviderIP && ip.Compare(newIP) != 0 {
+				return netip.Addr{}, fmt.Errorf("%w: sent ip %s to update but received %s",
+					errors.ErrIPReceivedMismatch, ip, newIP)
+			}
+			return newIP, nil
+		case strings.Contains(s, "nochg"):
+			fmt.Println(s)
```
qdm12, line 147:
```suggestion
```

internal/provider/providers/nowdns/provider.go
```diff
@@ -0,0 +1,171 @@
...
+			newIP, err = netip.ParseAddr(ip.String())
+			if err != nil {
+				return netip.Addr{}, fmt.Errorf("%w: in response %q", errors.ErrReceivedNoResult, s)
+			} else if !p.useProviderIP && ip.Compare(newIP) != 0 {
+				return netip.Addr{}, fmt.Errorf("%w: sent ip %s to update but received %s",
+					errors.ErrIPReceivedMismatch, ip, newIP)
+			}
+			return newIP, nil
```
qdm12, line 155: same here, this has to be changed 😉

internal/provider/providers/nowdns/provider.go
```diff
@@ -0,0 +1,171 @@
...
+			return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrUnknownResponse, s)
+		}
+	case http.StatusBadRequest:
+		switch s {
+		case constants.Nohost:
+			return netip.Addr{}, fmt.Errorf("%w", errors.ErrHostnameNotExists)
+		case constants.Badauth:
+			return netip.Addr{}, fmt.Errorf("%w", errors.ErrAuth)
+		default:
+			return netip.Addr{}, fmt.Errorf("%w: %d: %s", errors.ErrHTTPStatusNotValid, response.StatusCode, s)
```
qdm12, line 166: since we don't know how to handle the returned response, let's use the unknown response error instead

```suggestion
			return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrUnknownResponse, s)
```

qdm12 (changes requested):
Nice work, thanks 💯 !

Just a few minor changes to be made, thanks 😉

qdm12 (changes requested):
Also please add nowdns to the list in the readme.md, and add a document `docs/nowdns.md`

internal/provider/constants/providers.go
```diff
@@ -35,6 +35,7 @@ const (
 	Netcup       models.Provider = "netcup"
 	Njalla       models.Provider = "njalla"
 	NoIP         models.Provider = "noip"
+	NowDNS       models.Provider = "nowdns"
```
qdm12, line 38: Also add NowDNS to the `ProviderChoices` function below

internal/provider/providers/nowdns/provider.go
CyberAustin: Ditto. Replaced with `ip`

internal/provider/providers/nowdns/provider.go
CyberAustin: Corrected per your suggestion.

CyberAustin:
> Also please add nowdns to the list in the readme.md, and add a document `docs/nowdns.md`

Apparently in my sleep deprived state I neglected to add this to my last commit :grimacing:

go.mod
CyberAustin: I get past the linter now, but the image fails to build. Hits me with this:

`Step 37/50 : RUN GOARCH="$(xcputranslate translate -targetplatform ${TARGETPLATFORM} -field arch)"     GOARM="$(xcputranslate translate -targetplatform ${TARGETPLATFORM} -field arm)"     go build -trimpath -ldflags="-s -w     -X 'main.version=$VERSION'     -X 'main.created=$CREATED'     -X 'main.commit=$COMMIT'     " -o app cmd/updater/main.go
 ---> Running in 66d5b79a6e51
go: invalid GOARM: must be 5, 6, 7
The command '/bin/sh -c GOARCH="$(xcputranslate translate -targetplatform ${TARGETPLATFORM} -field arch)"     GOARM="$(xcputranslate translate -targetplatform ${TARGETPLATFORM} -field arm)"     go build -trimpath -ldflags="-s -w     -X 'main.version=$VERSION'     -X 'main.created=$CREATED'     -X 'main.commit=$COMMIT'     " -o app cmd/updater/main.go' returned a non-zero code: 2
`

CyberAustin:
Changes made. Docker build still failing though.

`Step 37/50 : RUN GOARCH="$(xcputranslate translate -targetplatform ${TARGETPLATFORM} -field arch)"     GOARM="$(xcputranslate translate -targetplatform ${TARGETPLATFORM} -field arm)"     go build -trimpath -ldflags="-s -w     -X 'main.version=$VERSION'     -X 'main.created=$CREATED'     -X 'main.commit=$COMMIT'     " -o app cmd/updater/main.go
 ---> Running in 5c6c1893196b
go: invalid GOARM: must be 5, 6, 7
The command '/bin/sh -c GOARCH="$(xcputranslate translate -targetplatform ${TARGETPLATFORM} -field arch)"     GOARM="$(xcputranslate translate -targetplatform ${TARGETPLATFORM} -field arm)"     go build -trimpath -ldflags="-s -w     -X 'main.version=$VERSION'     -X 'main.created=$CREATED'     -X 'main.commit=$COMMIT'     " -o app cmd/updater/main.go' returned a non-zero code: 2`

qdm12:
The build fails because of `go: invalid GOARM: must be 5, 6, 7` it might be a bug from xcputranslate. What's your host platform? You can do uname -a

CyberAustin:
Linux mint 5.15.0-76-generic #83-Ubuntu SMP Thu Jun 15 19:16:32 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

go.mod
CyberAustin: Moving to main thread.

CyberAustin:
Fixed the Dockerfile.

CyberAustin:
Okay, builds inside of a GitHub codespace container just fine with the original Dockerfile. Build output below.

```#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 32B done
#1 DONE 0.1s

#2 [internal] load .dockerignore
#2 transferring context: 35B done
#2 DONE 0.2s

#3 [internal] load metadata for docker.io/library/golang:1.20-alpine3.18
#3 DONE 0.5s

#4 [internal] load metadata for docker.io/qmcgaw/xcputranslate:v0.6.0
#4 DONE 0.5s

#5 [internal] load metadata for docker.io/qmcgaw/binpot:mockgen-v1.6.0
#5 DONE 0.5s

#6 [internal] load metadata for docker.io/qmcgaw/binpot:golangci-lint-v1.53.2
#6 DONE 0.6s

#7 [mockgen 1/1] FROM docker.io/qmcgaw/binpot:mockgen-v1.6.0@sha256:9338f2b706b42197fc5d46f02f03f8a93029b7bea9c819068cb2581f75d3d5b6
#7 DONE 0.0s

#8 [golangci-lint 1/1] FROM docker.io/qmcgaw/binpot:golangci-lint-v1.53.2@sha256:f2517d64e10011cbb98d9ae57f3ddd26d82237f139084767a001df162e1252a6
#8 DONE 0.0s

#9 [xcputranslate 1/1] FROM docker.io/qmcgaw/xcputranslate:v0.6.0@sha256:610582458e71facb54fd16cfb207e8b233afb6e15178742aa5332fd257926b74
#9 DONE 0.0s

#10 [base  1/11] FROM docker.io/library/golang:1.20-alpine3.18@sha256:7839c9f01b5502d7cb5198b2c032857023424470b3e31ae46a8261ffca72912a
#10 DONE 0.0s

#11 [internal] load build context
#11 transferring context: 18.88kB 0.0s done
#11 DONE 0.1s

#12 [base  9/11] COPY pkg/ ./pkg/
#12 CACHED

#13 [build 1/1] RUN GOARCH="$(xcputranslate translate -targetplatform linux/amd64 -field arch)"     GOARM="$(xcputranslate translate -targetplatform linux/amd64 -field arm)"     go build -trimpath -ldflags="-s -w     -X 'main.version=unknown'     -X 'main.created=an unknown date'     -X 'main.commit=unknown'     " -o app cmd/updater/main.go
#13 CACHED

#14 [base  4/11] COPY --from=xcputranslate /xcputranslate /usr/local/bin/xcputranslate
#14 CACHED

#15 [base  5/11] COPY --from=golangci-lint /bin /go/bin/golangci-lint
#15 CACHED

#16 [base  3/11] RUN apk --update add git g++ findutils
#16 CACHED

#17 [base  2/11] WORKDIR /tmp/gobuild
#17 CACHED

#18 [base  8/11] RUN go mod download
#18 CACHED

#19 [base 10/11] COPY cmd/ ./cmd/
#19 CACHED

#20 [base 11/11] COPY internal/ ./internal/
#20 CACHED

#21 [base  6/11] COPY --from=mockgen /bin /go/bin/mockgen
#21 CACHED

#22 [base  7/11] COPY go.mod go.sum ./
#22 CACHED

#23 [stage-8 1/1] COPY --from=build --chown=1000:1000 /tmp/gobuild/app /updater/app
#23 CACHED

#24 exporting to image
#24 exporting layers done
#24 writing image sha256:bececb724395b216a77f82f08132c1b7a9c0c75bdb63bf3073f328d9a8fc8661 0.0s done
#24 naming to docker.io/library/ddns-updater-blank done
#24 DONE 0.0s
WARNING: buildx: failed to read current commit information with git rev-parse --is-inside-work-tree
```

qdm12:
That's most likely because your local docker is a bit outdated or doesn't use buildkit/buildx, which should set TARGETPLATFORM to `linux/amd64` in your case automatically.

Now I'm trying to figure out how to prevent this issue arising again on such Docker setups.

First, there is a bug where TARGETPLATFORM cannot have a default in the Dockerfile: https://github.com/docker/buildx/issues/510 so setting it to `linux/amd64` won't work for cross compilation.

As it is, xcputranslate fails and exits with 1, for example `xcputranslate translate -targetplatform "" -field arch` produces

```
Fatal error: invalid platform: string is malformed: ""
exit status 1
```

But because it's wrapped in a subshell in `GOARCH="$(xcputranslate translate -targetplatform ${TARGETPLATFORM} -field arch)"`, the command is only silently failing and really GOARM is set to `Fatal error: invalid platform: string is malformed: ""`... Maybe you would have an idea how to fix the GOARCH and GOARM lines in the Dockerfile so it crashes if xcputranslate exits with 1??

There is also an alternative where xcputranslate handles the empty TARGETPLATFORM string and outputs the current platform, but that's a bit of a pain to implement and reliably detect the platform version, so I would rather avoid.

docs/nowdns.md
```diff
@@ -0,0 +1,26 @@
...
+      "password": "password",
+    }
+  ]
+}
+```
+
+### Compulsory parameters
+
+- `"domain"` your domain name
+- `"host"` leave it set to `"@"`
```
qdm12, line 24: If the host can only be `"@"`, should we just remove the option then?

docs/nowdns.md
```diff
@@ -0,0 +1,26 @@
...
+      "username": "username",
+      "password": "password",
+    }
+  ]
+}
+```
+
+### Compulsory parameters
+
+- `"domain"` your domain name
```
qdm12, line 23: nit unnecessary description

```suggestion
- `"domain"`
```

internal/provider/providers/nowdns/provider.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/qdm12/ddns-updater/internal/provider/constants"
+	"github.com/qdm12/ddns-updater/internal/provider/errors"
+	"github.com/qdm12/ddns-updater/internal/provider/headers"
+	"github.com/qdm12/ddns-updater/internal/provider/utils"
+	"github.com/qdm12/ddns-updater/pkg/publicip/ipversion"
+)
+
+type Provider struct {
+	domain        string
+	host          string
```
qdm12, line 23: you can remove the host field if it can only ever be `"@"`

internal/provider/providers/nowdns/provider.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func (p *Provider) isValid() error {
+	switch {
+	case p.username == "":
+		return fmt.Errorf("%w", errors.ErrUsernameNotSet)
+	case p.password == "":
+		return fmt.Errorf("%w", errors.ErrPasswordNotSet)
+	case p.host == "*":
+		return fmt.Errorf("%w", errors.ErrHostWildcard)
```
qdm12, line 63: In the document, you mention to leave the host to `@`, so can it be something else than `@`? 🤔

CyberAustin:
But is that a bug with this repo, or something upstream?

docs/nowdns.md
CyberAustin: I did that and it was failing to update the record.

docs/nowdns.md
CyberAustin: I felt that over documentation was better than under.

internal/provider/providers/nowdns/provider.go
CyberAustin: Fair point. Fixed.

internal/provider/providers/nowdns/provider.go
CyberAustin: Removed.

CyberAustin: It has to be passed as "@", otherwise the API throws a "nohost" error.

qdm12 (changes requested):
> I did that and it was failing to update the record.
> It has to be passed as "@", otherwise the API throws a "nohost" error.
> Fair point. Fixed.

If the host can only be `@` (not `*` or `abc`), then remove the option `host` from the config and the field `host` in the provider struct.

CyberAustin:
But without host field, it fails. "nohost" is what the API returns.

internal/provider/providers/nowdns/provider.go
```diff
@@ -0,0 +1,167 @@
...
+func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Addr) (newIP netip.Addr, err error) {
+	u := url.URL{
+		Scheme: "https",
+		Host:   "now-dns.com",
+		Path:   "/update",
+		User:   url.UserPassword(p.username, p.password),
+	}
+
+	values := url.Values{}
+	values.Set("hostname", utils.BuildURLQueryHostname(p.host, p.domain))
```
qdm12, line 108: Just have this as `p.domain`, so it won't be as `.domain.com` if the host is empty.

```suggestion
	values.Set("hostname", p.domain)
```

qdm12 (changes requested):

CyberAustin:
> > I did that and it was failing to update the record.
> > It has to be passed as "@", otherwise the API throws a "nohost" error.
> > Fair point. Fixed.
> 
> If the host can only be `@` (not `*` or `abc`), then remove the option `host` from the config and the field `host` in the provider struct.

This opens up a whole lot of other issues, like having to refactor host out of the rest of the file, redefine or replace utils.ToString(p.domain, p.host, constants.NowDNS, p.ipVersion) as a variadatic function. I think I found a simpler solution that will be in the next commit that I think will work just as well.

CyberAustin:
Basically just set `host` to `@` in the provider.go file so that it is always that way, and removed any reference to host in the documentation.

internal/provider/providers/nowdns/provider.go
CyberAustin: Vicariously fixed this. See below conversation and commit https://github.com/qdm12/ddns-updater/pull/504/commits/e6bb0f37dd4b76b33036e73edce4297a99083afd

CyberAustin (commented):
I think the solution is good, simple and ready to merge into master.

qdm12 (approved):
Awesome, great work 💯  Merging it 😉 

side note: I went ahead and removed the host injection and host field in the struct (hardcoding `@` where needed), since that was itching my brain too much to have a constant struct field and unused argument injected.

PR #513, opened 2023-08-04
Fix IPs in some logging messages

felixwrt:
I noticed a small error in some of the logging messages. Most likely a Copy&Paste error.


qdm12 (approved):
Totally my bad, great work 💯 !!

PR #515, opened 2023-08-07
fix small error in service.go

felixwrt:
Just saw this small issue (most likely another copy&paste error). Using `ip` instead of `ipv6` here doesn't make sense.


qdm12 (approved):
Totally indeed, thanks so much 👍 That will resolve some headaches!

PR #539, opened 2023-10-07
feat(provider): add support for Mikrotik routers

jhotmann:
When setting up hairpin NAT on Mikrotik routers, it is much easier if you have your public IP set in an [address list](https://wiki.mikrotik.com/wiki/Manual:IP/Firewall/Address_list). This provider adds the ability to update an address list via the router's API.


qdm12 (commented):
Hi there! I rebased your branch and improved/modified a bit. I'm also about to work into the code to remove the routeros dependency; I have a few questions:

1. Why is IPv6 not allowed? I would tend to think it should work as well as IPv4 right?
2. Why limit to one item in the address list? Is it not possible to handle multiple? I changed the code to handle multiple, but let me know if that's a problem.

Thanks!

qdm12:
Please try the latest code from your branch. I pushed another commit to remove the dependency on routeros (re-coding the logic needed), saving ~100KB of program size 😄

qdm12:
Up? Does it work now?

PR #568, opened 2023-12-10
Gandi Personal Access Token support

varner-owl:
Currently authentication for updates to Gandi domains can only be done with the Gandi API key. [Gandi has depreciated the API key in favor of Personal Access Tokens. 
](https://api.gandi.net/docs/authentication/)
This PR adds the option to authenticate using a Personal Access Tokens but does not require a token to be used instead of an API key. This should not break anyone's current configurations or require updates to their configuration files. This implementation produces a KeyNotSet error only if both the API Key and Personal Access Token are not provided or are empty the config file. If both an API key and Personal Access Token are provided, it will default to using the Token for authentication. It does not use the API key as a fall back (i.e. if authentication via Token fails, try again with the API key), but this could be implemented in the future if desired.

Example config.json with different possible valid configurations:
```
{
    "settings": [
        {
            "provider": "gandi",
            "domain": "mydomain.com",
            "host": "@",
            "key": "mysecretkey",
            "ip_version": "ipv4"
        },
        {
            "provider": "gandi",
            "domain": "mydomain.com",
            "host": "@",
            "token": "mypersonalaccesstoken",
            "ip_version": "ipv4"
        },
        {
            "provider": "gandi",
            "domain": "mydomain.com",
            "host": "@",
            "key": "mysecretkey",
            "token": "mypersonalaccesstoken",
            "ip_version": "ipv4"
        }
    ]
}
```

### DIFF
--- a/docs/gandi.md
+++ b/docs/gandi.md
@@ -13,8 +13,7 @@ This provider uses Gandi v5 API
       "provider": "gandi",
       "domain": "domain.com",
       "host": "@",
-      "key": "key",
-      "token": "token",
+      "personal_access_token": "token",
       "ttl": 3600,
       "ip_version": "ipv4"
     }
@@ -26,8 +25,7 @@ This provider uses Gandi v5 API
 
 - `"domain"`
 - `"host"` which can be a subdomain, `@` or a wildcard `*`
-- Either a Gandi API Key `"key"` or Gandi Personal Access Token `"token"` must be provided.
-- If both a `"key"` and `"token"` are provided, the `"token"` will be used.
+- `"personal_access_token"`
 
 ### Optional parameters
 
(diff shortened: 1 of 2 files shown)


Trambert:
Hello,
This is becoming urgent, Gandi's API endpoint deprecated the use of the API key today.
Thank you very much for your work!

qdm12:
> This is becoming urgent, Gandi's API endpoint deprecated the use of the API key today.

By "deprecated", it should still work right? It's just that the PAT is now the preferred method right?
I reviewed the code and made changes, I'll merge this in a few minutes anyway.

Notable changes I did:

- Change `token` -> `personal_access_token` in the JSON, to avoid any confusion regarding API key vs personal access token.
- Remove `key` from documentation docs/gandi.md, although it still works

varner-owl:
> > This is becoming urgent, Gandi's API endpoint deprecated the use of the API key today.
> 
> By "deprecated", it should still work right? It's just that the PAT is now the preferred method right? I reviewed the code and made changes, I'll merge this in a few minutes anyway.

It was also my understanding that PAT is just the preferred method now. Although I stopped using the API key last month, I just checked again and the API keys still work and Gandi has not updated their documentation to say anything about the removal of API keys. Either way, thanks for merging it and for all the work you do on all these repositories!

PR #634, opened 2024-02-05
Support named volumes

yorickdowne:
Create an /updater/data directory owned by the app's user so that a named volume mapped into that location has the right permissions

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,7 @@ RUN git init && \
     rm -rf .git/
 
 FROM --platform=$BUILDPLATFORM base AS build
+RUN mkdir -p /tmp/data
 ARG VERSION=unknown
 ARG CREATED="an unknown date"
 ARG COMMIT=unknown
@@ -69,6 +70,7 @@ ARG UID=1000
 ARG GID=1000
 USER ${UID}:${GID}
 ENTRYPOINT ["/updater/app"]
+COPY --from=build --chown=${UID}:${GID} /tmp/data /updater/data
 ENV \
     # Core
     CONFIG= \


Dockerfile
```diff
@@ -61,6 +61,7 @@ RUN GOARCH="$(xcputranslate translate -targetplatform ${TARGETPLATFORM} -field a
     -X 'main.date=$CREATED' \
     -X 'main.commit=$COMMIT' \
     " -o app cmd/updater/main.go
+RUN mkdir -p /tmp/data
```
qdm12, line 64: Very minor detail/optimization, can you move this line right after `FROM --platform=$BUILDPLATFORM base AS build`? That way it isn't executed on every code change/rebuilds (assuming layer caching). Not a big deal though 😉

Dockerfile
```diff
@@ -112,3 +113,4 @@ LABEL \
     org.opencontainers.image.title="ddns-updater" \
     org.opencontainers.image.description="Universal DNS updater with WebUI"
 COPY --from=build --chown=${UID}:${GID} /tmp/gobuild/app /updater/app
+COPY --from=build --chown=${UID}:${GID} /tmp/data /updater/data
```
qdm12, line 116: Same here, can you move this line right after `ENTRYPOINT ["/updater/app"]` since it shouldn't change really (unlike `ENV` / `ARG` etc.)

qdm12 (approved):
Nice suggestion! 💯 
Just two super minor comments to move lines around, but otherwise good to merge!

Dockerfile
yorickdowne: Done

qdm12 (approved):

qdm12:
Merged, thanks again 💯 !

PR #638, opened 2024-02-07
Implementated global validation of domain strings

### LINKED ISSUE #626: Maintenance: validate domain strings for all providers
All domain strings are not validated for now. It would be good to check at a global scope the domain is

- empty
- a valid fqdn (without trail dot)

Some providers only use a host and the domain json value can be the empty string (constant value in the code), for example duckdns the domain is always duckdns.org. goip.de can have the domain empty, which defaults to goip.de, but it can also be set for example to goip.it.

CyberAustin:
Should close #626 and provide a basic sanity check for all providers, agnostic of provider specific requirements.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,6 +21,7 @@ require (
 require (
 	cloud.google.com/go/compute v1.19.1 // indirect
 	cloud.google.com/go/compute/metadata v0.2.3 // indirect
+	github.com/chmike/domain v1.0.1 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/fatih/color v1.15.0 // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.21
 
 require (
 	github.com/breml/rootcerts v0.2.16
+	github.com/chmike/domain v1.0.1
 	github.com/containrrr/shoutrrr v0.8.0
 	github.com/go-chi/chi/v5 v5.0.11
 	github.com/golang/mock v1.6.0
@@ -21,7 +22,6 @@ require (
 require (
 	cloud.google.com/go/compute v1.19.1 // indirect
 	cloud.google.com/go/compute/metadata v0.2.3 // indirect
-	github.com/chmike/domain v1.0.1 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/fatih/color v1.15.0 // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
--- a/go.sum
+++ b/go.sum
@@ -10,6 +10,8 @@ github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03
 github.com/breml/rootcerts v0.2.16 h1:yN1TGvicfHx8dKz3OQRIrx/5nE/iN3XT1ibqGbd6urc=
 github.com/breml/rootcerts v0.2.16/go.mod h1:S/PKh+4d1HUn4HQovEB8hPJZO6pUZYrIhmXBhsegfXw=
 github.com/census-instrumentation/opencensus-proto v0.2.1/go.mod h1:f6KPmirojxKA12rnyqOA5BBL4O983OfeGPqjHWSTneU=
+github.com/chmike/domain v1.0.1 h1:ug6h3a7LLAfAecBAysbCXWxP1Jo8iBKWNVDxcs1BNzA=
+github.com/chmike/domain v1.0.1/go.mod h1:h558M2qGKpYRUxHHNyey6puvXkZBjvjmseOla/d1VGQ=
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
 github.com/cncf/udpa/go v0.0.0-20191209042840-269d4d468f6f/go.mod h1:M8M6+tZqaGXZJjfX53e64911xZQV5JYwmTeXPW+k8Sc=
 github.com/containrrr/shoutrrr v0.8.0 h1:mfG2ATzIS7NR2Ec6XL+xyoHzN97H8WPjir8aYzJUSec=
--- a/internal/params/json.go
+++ b/internal/params/json.go
@@ -8,7 +8,6 @@ import (
 	"io/fs"
 	"net/netip"
 	"os"
-  "reflect"
 	"strings"
 
 	"github.com/qdm12/ddns-updater/internal/models"
@@ -152,27 +151,22 @@ func makeSettingsFromObject(common commonSettings, rawSettings json.RawMessage,
 	if common.Provider == "google" {
 		return nil, nil, fmt.Errorf("%w: %s", ErrProviderNoLongerSupported, common.Provider)
 	}
-  if ((common.Host != "" && common.Host != "@") && common.Domain != "") {
-    FQDN := common.Host + "." + common.Domain
-    fmt.Println(reflect.TypeOf(common.Host))
-    fmt.Println(len(common.Host))
-    err := domain.Check(FQDN)
-		if err != nil {
-      return nil, nil, fmt.Errorf("%w: %s", errFQDNInvalid, FQDN)
-      }
-	} else if (common.Host == "@" || common.Host == "") && common.Domain != "" {
-      err := domain.Check(common.Domain)
-      if err != nil {
-        return nil, nil, fmt.Errorf("%w: %s", errFQDNInvalid, common.Domain)
-    } 
-  } else if common.Host != "" && common.Host != "@" && common.Domain == "" {
-      FQDN := common.Host + ".test.com"
-      err := domain.Check(FQDN)
-      if err != nil {
-        return nil, nil, fmt.Errorf("%w: %s", errFQDNInvalid, FQDN)
-    }
+  
+  FQDN := ""
+  switch {
+    case (common.Host != "" && common.Host != "@") && common.Domain != "":
+      FQDN = common.Host + "." + common.Domain
+    case (common.Host == "@" || common.Host == "") && common.Domain != "":
+      FQDN = common.Domain
+    case common.Host != "" && common.Host != "@" && common.Domain == "":
+      FQDN = common.Host + ".test.com"
   }
-
+  
+  err = domain.Check(FQDN)
+  if err != nil {
+    return nil, nil, fmt.Errorf("%w: %s", errFQDNInvalid, FQDN)
+  }
+  
 	providerName := models.Provider(common.Provider)
 	if providerName == constants.DuckDNS { // only hosts, no domain
 		if common.Domain != "" { // retro compatibility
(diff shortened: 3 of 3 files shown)


internal/params/json.go
```diff
@@ -149,7 +152,18 @@ func makeSettingsFromObject(common commonSettings, rawSettings json.RawMessage,
...
+	match, _ := regexp.MatchString(FQDNRegex, FQDN)
+
+	if common.Host == "" || common.Domain == "" {
+		return nil, nil, fmt.Errorf("%w", errHostDomainBlank)
+	} else if !match {
+		return nil, nil, fmt.Errorf("%w: %s", errFQDNInvalid, FQDN)
+	}
+
 	providerName := models.Provider(common.Provider)
+
```
qdm12, line 166: nit unneeded change

```suggestion
```

internal/params/json.go
```diff
@@ -149,7 +152,18 @@ func makeSettingsFromObject(common commonSettings, rawSettings json.RawMessage,
 		return nil, nil, fmt.Errorf("%w: %s", ErrProviderNoLongerSupported, common.Provider)
 	}
 
+	FQDNRegex := "^(?i)(@?|[a-z0-9-]+)(\\.[a-z0-9-]{2,})+$"
+	FQDN := (common.Host + "." + common.Domain)
+	match, _ := regexp.MatchString(FQDNRegex, FQDN)
```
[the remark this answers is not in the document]
CyberAustin: But we're not always using RFC compliant FQDNs. For example, when host is set to "@", this module marks it as invalid: https://go.dev/play/p/PiRXqaS4wdE

internal/params/json.go
```diff
@@ -149,7 +152,18 @@ func makeSettingsFromObject(common commonSettings, rawSettings json.RawMessage,
...
+
+	if common.Host == "" || common.Domain == "" {
```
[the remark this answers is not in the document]
CyberAustin: > The domain can be empty (see issue for examples mentioned). 

Okay, but in #626 you said "All domain strings are not validated for now. It would be good to check at a global scope the domain is: empty" 😂

internal/params/json.go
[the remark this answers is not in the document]
CyberAustin: Plus, to account for all these edge cases, it's going to add like 3 or 4 nested if/ifelse statements, which are all simply covered under this one regex statement.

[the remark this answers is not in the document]
CyberAustin: This is what I'm currently working to implement chmike/domain. I feel like the regex is cleaner.

```
if ((common.Host != "" || common.Host != "@") && common.Domain != "") {
    FQDN := common.Host + "." + common.Domain
    fmt.Println(reflect.TypeOf(common.Host))
    fmt.Println(len(common.Host))
    err := domain.Check(FQDN)
		if err != nil {
      return nil, nil, fmt.Errorf("%w: %s", errFQDNInvalid, FQDN)
      }
  } else if (common.Host == "@" || common.Host == "") && common.Domain != "" {
      err := domain.Check(common.Domain)
      if err != nil {
        return nil, nil, fmt.Errorf("%w: %s", errFQDNInvalid, common.Domain)
    } 
  } else if common.Host != "" && common.Host != "@" && common.Domain == "" {
      FQDN := common.Host + ".test.com"
      err := domain.Check(FQDN)
      if err != nil {
        return nil, nil, fmt.Errorf("%w: %s", errFQDNInvalid, FQDN)
    }
  }
```

CyberAustin:
Ugh. Can't run linters or devcontainer in Replit. Have to fix this later at home.

CyberAustin:
I don't know why the linter is failing in GitHub Actions because it passes in the dev container. Anyway, ready for review.

internal/params/json.go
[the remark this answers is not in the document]
qdm12: Don't validate the host though, only the domain.
We could validate the host eventually (although it has to be validated again for each provider using it with restriction), but in another PR 🙏

internal/params/json.go
```diff
@@ -14,6 +14,8 @@ import (
 	"github.com/qdm12/ddns-updater/internal/provider"
 	"github.com/qdm12/ddns-updater/internal/provider/constants"
 	"github.com/qdm12/ddns-updater/pkg/publicip/ipversion"
+  "github.com/chmike/domain"
+
```
qdm12, line 18: indentation/formatting seems a bit wrong, that's probably why the linter is failing 🤔

internal/params/json.go
```diff
@@ -148,7 +151,22 @@ func makeSettingsFromObject(common commonSettings, rawSettings json.RawMessage,
...
+  
+  FQDN := ""
+  switch {
+    case (common.Host != "" && common.Host != "@") && common.Domain != "":
+      FQDN = common.Host + "." + common.Domain
+    case (common.Host == "@" || common.Host == "") && common.Domain != "":
+      FQDN = common.Domain
+    case common.Host != "" && common.Host != "@" && common.Domain == "":
+      FQDN = common.Host + ".fakedomain.com"
+  }
```
qdm12, line 163: Let's focus on validating the domain field for now only (what the PR and issue are for).
The host can be `@`, `*` or a valid subdomain (I'll hunt for an RFC precising what's the allowed characters etc.), but let's do this in a subsequent PR.

internal/params/json.go
```diff
@@ -148,7 +151,22 @@ func makeSettingsFromObject(common commonSettings, rawSettings json.RawMessage,
...
+  
+  err = domain.Check(FQDN)
+  if err != nil {
+    return nil, nil, fmt.Errorf("%w: %s", errFQDNInvalid, FQDN)
```
qdm12, line 167:
```suggestion
    return nil, nil, fmt.Errorf("validating domain: %w", err)
```

go.mod
```diff
@@ -21,6 +21,7 @@ require (
 require (
 	cloud.google.com/go/compute v1.19.1 // indirect
 	cloud.google.com/go/compute/metadata v0.2.3 // indirect
+	github.com/chmike/domain v1.0.1 // indirect
```
qdm12, line 24: You might want to run `go mod tidy` to move this to a direct dependency instead of indirect.

internal/params/json.go
CyberAustin: The now included module does this well now that I've poked at it, just need to account for these special cases where the host won't RFC compliant.

internal/params/json.go
```diff
@@ -138,17 +139,20 @@ func extractAllSettings(jsonBytes []byte) (
 	return allProviders, warnings, nil
 }
 
-var (
-	ErrProviderNoLongerSupported = errors.New("provider no longer supported")
-)
+var ErrProviderNoLongerSupported = errors.New("provider no longer supported")
```
qdm12, line 142: nit revert, this is an unneeded change

go.mod
```diff
@@ -3,7 +3,8 @@ module github.com/qdm12/ddns-updater
 go 1.21
 
 require (
-	github.com/breml/rootcerts v0.2.16
+	github.com/breml/rootcerts v0.2.15
```
qdm12, line 6: nit can you rebase your branch on the master branch 🙏 Thanks!

qdm12 (commented):
Nice 👍 💯 

Now, we need to add a check the domain isn't `""` for the providers actually using the domain parameter. Most of them do, but there are 2-3 exceptions (like duckdns, goip). So that would be edits in each of the provider subpackage within `internal/provider/providers`

internal/params/json.go
CyberAustin: Odd. I had fixed that already.

CyberAustin:
> Now, we need to add a check the domain isn't `""` for the providers actually using the domain parameter. Most of them do, but there are 2-3 exceptions (like duckdns, goip). So that would be edits in each of the provider subpackage within `internal/provider/providers`

I'm not sure what you're asking here. Could you explain it a different way.

qdm12:
We need to add a check in most providers that the domain is not empty `""`. Because in the current "above layer" check, we check the domain is empty (🆗) or valid (🆗), but some providers require a non-empty domain (most except a few).

For example in https://github.com/qdm12/ddns-updater/blob/6a6b1a8ebbaf228f3454d6a1f4a2a61addea6c0d/internal/provider/providers/cloudflare/provider.go#L97

add a case to check the domain is not empty:

```go
case p.domain == "":
    return fmt.Errorf("%w", errors.ErrDomainNotSet)
```

CyberAustin:
Should be good now. Please take a look @qdm12

go.sum
```diff
@@ -7,9 +7,11 @@ cloud.google.com/go/compute/metadata v0.2.3/go.mod h1:VAV5nSsACxMJvgaAuX6Pk2Aawl
 cloud.google.com/go/longrunning v0.4.1 h1:v+yFJOfKC3yZdY6ZUI933pIYdhyhV8S3NpWrXWmg7jM=
 cloud.google.com/go/longrunning v0.4.1/go.mod h1:4iWDqhBZ70CvZ6BfETbvam3T8FMvLK+eFj0E6AaRQTo=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/breml/rootcerts v0.2.16 h1:yN1TGvicfHx8dKz3OQRIrx/5nE/iN3XT1ibqGbd6urc=
-github.com/breml/rootcerts v0.2.16/go.mod h1:S/PKh+4d1HUn4HQovEB8hPJZO6pUZYrIhmXBhsegfXw=
+github.com/breml/rootcerts v0.2.15 h1:zLq1U/uZIJJVZDqCHfeE5l+JLx/U+F3rGyi7wZcFmJA=
+github.com/breml/rootcerts v0.2.15/go.mod h1:S/PKh+4d1HUn4HQovEB8hPJZO6pUZYrIhmXBhsegfXw=
```
qdm12, line 11: please rebase on the master branch to remove this diffs (reverting to previous version)

qdm12 (commented):
Thanks!
Positive is it does it in 3 lines for all providers except the two weird ones
Negative side is it it adds coupling in the code, but oh well that's fine this time 😉 

Just rebase/fix the go.mod dependency diff and I'll merge, thanks! 🎖️

go.sum
CyberAustin: Done. Back to you.

go.mod
qdm12, line 6: It's still there 😢
You can just go get the version 0.2.16 and commit it I think (anyway it'll get squashed)

CyberAustin: What 😡 I'm clearly doing something wrong I guess. Let me look at it again.

CyberAustin: Well if that didn't work, then I don't know what to do. Everything I'm seeing for the `global_val` branch is 0.2.15 🤷

internal/params/json.go
```diff
@@ -149,6 +151,17 @@ func makeSettingsFromObject(common commonSettings, rawSettings json.RawMessage,
 		return nil, nil, fmt.Errorf("%w: %s", ErrProviderNoLongerSupported, common.Provider)
 	}
 
+	if common.Domain == "" && (common.Provider != "duckdns" && common.Provider != "goip") {
+		return nil, nil, fmt.Errorf("%w: %s", ErrDomainBlank, common.Provider)
```
qdm12, line 155:
```suggestion
		return nil, nil, fmt.Errorf("%w: for provider %s", ErrDomainBlank, common.Provider)
```

qdm12 (approved):

PR #696, opened 2024-04-19
Update 'remote containers' references to the new name 'Dev Containers'

ryankupk:
Very minor update to the documents for contributing to use the new name of the remote containers extension: dev containers


qdm12 (approved):
Thanks! Just holding before merging so I can update the documentation across my repositories (they all more or less share the same readme)

Issue #375, opened 2022-10-11
Feature request: AWS Route 53


pbarone:
1. What's the feature?
Support for AWS Route 53

2. Extra information?

willsargent:
Bump!  This would be amazing!

roseap42:
Agree. I'm looking for a new all-in-one ddns updater, and have a route53 domain that needs to be updated. Great project you've got here!

seth586:
Agreed, would appreciate!

qdm12:
You can build the (beta route53) image with docker and git using:

```sh
docker build -t qmcgaw/ddns-updater https://github.com/marcelohpf/ddns-updater.git#feat/aws-simple-record
```

And then try it following https://github.com/marcelohpf/ddns-updater/blob/feat/aws-simple-record/docs/route53.md

I'm still waiting for feedback to check it works fine, please reply on https://github.com/qdm12/ddns-updater/pull/715 thanks!

heisian:
@qdm12 Thank you for your work on this - excited to use it once it's out of beta!

qdm12:
Whoops sorry I forgot to close the issue this in the latest and v2.7.0 release 😉 See the readme

PR #715, opened 2024-05-09
feat(aws): add support to route53 simple records

marcelohpf:
# Changes

This includes the support to Simple Route policy from Route53. 

# Details:
* Set default TTL to 300s as it is same default interval as ddns-updater
* It implements only Simple Routing, there are other 7 [policies available for route53](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-policy.html#routing-policy-geo), but it is not a use case for me. If there are larger interest on them, I can send more PRs.
* Providing the basic policy from documentation is enough for upserting domains following least privilege accesses
* Uses only Static Credentials following this project design (/data/config.json for all provider configs).
* It doesn't wait for propagation (GetChange API) because it updates one domain at a time and runs quite often.

# Testing

Build container and ran with the default configs and using policy in the documentation

Here is one event for a domain that didn't exit at route53

```json
{
    "eventVersion": "1.09",
    "userIdentity": {...},
    "eventTime": "2024-05-09T18:10:47Z",
    "eventSource": "route53.amazonaws.com",
    "eventName": "ChangeResourceRecordSets",
    "awsRegion": "us-east-1",
    "sourceIPAddress": "<redacted>",
    "userAgent": "DDNS-Updater quentin.mcgaw@gmail.com",
    "requestParameters": {
        "hostedZoneId": "<redacted>",
        "changeBatch": {
            "comment": "Record updated by ddns-updater",
            "changes": [
                {
                    "action": "UPSERT",
                    "resourceRecordSet": {
                        "name": "test.<my-domain>",
                        "type": "A",
                        "tTL": 300,
                        "resourceRecords": [
                            {
                                "value": "<redacted>"
                            }
                        ]
                    }
                }
            ]
        }
    },
    "responseElements": {
        "changeInfo": {
            "id": "/change/<redacted>",
            "status": "PENDING",
            "submittedAt": "May 9, 2024 6:10:47 PM",
            "comment": "Record updated by ddns-updater"
        }
    },
   ...
}
```

Note: this is a partial implementation of #375 because AWS route53 supports many more DNS features.


internal/provider/providers/aws/provider.go
```diff
@@ -0,0 +1,184 @@
...
+	err = json.Unmarshal(data, &providerSpecificSettings)
+	if err != nil {
+		return nil, fmt.Errorf("decoding provider specific settings: %w", err)
+	}
+
+	err = validateSettings(providerSpecificSettings, domain, host)
+	if err != nil {
+		return nil, fmt.Errorf("validating provider specific settings: %w", err)
+	}
+	ttl := int64(300)
```
qdm12, line 48: super-nit tll for DNS are int32 as far as I know 😉

internal/provider/providers/aws/provider.go
```diff
@@ -0,0 +1,184 @@
...
+	if providerSpecificSettings.TTL != nil {
+		ttl = *providerSpecificSettings.TTL
+	}
+
+	region := "us-east-1"
+	if providerSpecificSettings.Region != "" {
+		region = providerSpecificSettings.Region
+	}
```
qdm12, line 56: should we enforce the region to be set perhaps? What's the reason to default to `us-east-1`?

internal/provider/providers/aws/provider.go
```diff
@@ -0,0 +1,184 @@
...
+	HostedZoneID       string `json:"hosted_zone_id"`
+	TTL                *int64 `json:"ttl"`
+}
+
+func validateSettings(providerSpecificSettings settings, domain, host string) error {
+	// TODO: update this switch to be as restrictive as possible
+	// to fail early for the user. Use errors already defined
+	// in the internal/provider/errors package, or add your own
+	// if really necessary. When returning an error, always use
+	// fmt.Errorf (to enforce the caller to use errors.Is()).
```
qdm12, line 84:
```suggestion
```

internal/provider/providers/aws/provider.go
```diff
@@ -0,0 +1,184 @@
...
+
+func (p *Provider) BuildDomainName() string {
+	return utils.BuildDomainName(p.host, p.domain)
+}
+
+func (p *Provider) HTML() models.HTMLRow {
+	return models.HTMLRow{
+		Domain:    fmt.Sprintf("<a href=\"http://%s\">%s</a>", p.BuildDomainName(), p.BuildDomainName()),
+		Host:      p.Host(),
+		Provider:  "<a href=\"https://aws.amazon.com/route53/\">Amazon Route 53</a>",
```
qdm12, line 132: Should we call the provider `route53` instead of `aws`? 🤔

internal/provider/providers/aws/provider.go
```diff
@@ -0,0 +1,184 @@
...
+		IPVersion: p.ipVersion.String(),
+	}
+}
+
+func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Addr) (newIP netip.Addr, err error) {
```
qdm12, line 137: Is there an option in the AWS SDK to use the given `client`? This is somehow important for the rest of the program

internal/provider/providers/aws/provider.go
```diff
@@ -0,0 +1,184 @@
...
+	)
+	if err != nil {
+		return netip.Addr{}, fmt.Errorf("failed loading default config %w", err)
+	}
+
+	// TODO: implement option to wait for propagation
+	svcRoute53 := route53.NewFromConfig(cfg)
+	_, err = svcRoute53.ChangeResourceRecordSets(ctx, input)
+	if err != nil {
+		return netip.Addr{}, fmt.Errorf("%w", err)
```
qdm12, line 181:
```suggestion
		return netip.Addr{}, fmt.Errorf("changing resource record sets: %w", err)
```

internal/provider/providers/aws/provider.go
```diff
@@ -0,0 +1,184 @@
...
+		},
+		HostedZoneId: aws.String(p.hostedZoneID),
+	}
+
+	cfg, err := config.LoadDefaultConfig(ctx,
+		config.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(p.awsAccessKeyID, p.awsSecretAccessKey, "")),
+		config.WithRegion(p.region),
+	)
+	if err != nil {
+		return netip.Addr{}, fmt.Errorf("failed loading default config %w", err)
```
qdm12, line 174: nit don't prefix error wrappings with `failed to`  or `error ...`, since with multiple levels it repeats itself i.e. `failed to updated: failed to do this: failed to do that: actual error`

```suggestion
		return netip.Addr{}, fmt.Errorf("loading default config %w", err)
```

internal/provider/providers/aws/provider.go
```diff
@@ -0,0 +1,184 @@
...
+							{
+								Value: aws.String(ip.String()),
+							},
+						},
+						TTL: aws.Int64(p.ttl),
+						// SetIdentifier not set means Simple Routing,
+					},
+				},
+			},
+			Comment: aws.String("Record updated by ddns-updater"),
```
qdm12, line 164: Maybe to identify it better? Is this for Amazon to be aware of it, or as a personal note/comment?

```suggestion
			Comment: aws.String("Record updated by qdm12/ddns-updater"),
```

internal/provider/providers/aws/provider.go
```diff
@@ -0,0 +1,184 @@
...
+		config.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(p.awsAccessKeyID, p.awsSecretAccessKey, "")),
+		config.WithRegion(p.region),
+	)
+	if err != nil {
+		return netip.Addr{}, fmt.Errorf("failed loading default config %w", err)
+	}
+
+	// TODO: implement option to wait for propagation
+	svcRoute53 := route53.NewFromConfig(cfg)
+	_, err = svcRoute53.ChangeResourceRecordSets(ctx, input)
```
qdm12, line 179: Since this brings a lot of dependencies (I am trying to downscale on dependencies), do you think this would be possible to implement without the AWS SDK? I haven't looked into it yet personally, but I'm working on replacing the google sdk locally 😄

docs/aws.md
```diff
@@ -0,0 +1,67 @@
...
+  ]
+}
+```
+
+### Compulsory parameters
+
+- `"domain"`
+- `"host"` is your host and can be a subdomain or `"@"` or the wildcard `"*"`
+- `"aws_access_key_id"`
+- `"aws_secret_access_key"`
```
qdm12, line 33: I think we can remove the `aws_` prefix, it's obvious it's for aws/route53. I would also eventually rename those two to:



```suggestion
- `"key_id"`
- `"key"`
```

docs/aws.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Optional parameters
+
+- `"ip_version"` can be `ipv4` (A records), or `ipv6` (AAAA records) or `ipv4 or ipv6` (update one of the two, depending on the public ip found). It defaults to `ipv4 or ipv6`.
+- `"ipv6_suffix"` is the IPv6 interface identifiersuffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw public IPv6 address obtained is used in the record updating.
+- `"ttl"` amount of time, in seconds, that you want DNS recursive resolvers to cache information about this record. Defaults to `300`.
+- `"region"` of the route53 API. Route53 is a global resource, records created here will be globaly available unless you are using geolocation routing policy. Defaults to `us-east-1`.
+
+<!-- UPDATE THIS IF NEEDED -->
+
```
qdm12, line 46:
```suggestion
```

docs/aws.md
```diff
@@ -0,0 +1,67 @@
...
+### Compulsory parameters
+
+- `"domain"`
+- `"host"` is your host and can be a subdomain or `"@"` or the wildcard `"*"`
+- `"aws_access_key_id"`
+- `"aws_secret_access_key"`
+- `"hosted_zone_id"`
+
+<!-- UPDATE THIS IF NEEDED -->
```
qdm12, line 36:
```suggestion
```

docs/aws.md
```diff
@@ -0,0 +1,67 @@
...
+- `"ip_version"` can be `ipv4` (A records), or `ipv6` (AAAA records) or `ipv4 or ipv6` (update one of the two, depending on the public ip found). It defaults to `ipv4 or ipv6`.
+- `"ipv6_suffix"` is the IPv6 interface identifiersuffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw public IPv6 address obtained is used in the record updating.
+- `"ttl"` amount of time, in seconds, that you want DNS recursive resolvers to cache information about this record. Defaults to `300`.
+- `"region"` of the route53 API. Route53 is a global resource, records created here will be globaly available unless you are using geolocation routing policy. Defaults to `us-east-1`.
+
+<!-- UPDATE THIS IF NEEDED -->
+
+## Domain setup
+
+<!-- FILL THIS UP WITH A FEW NUMBERED STEPS -->
```
qdm12, line 49:
```suggestion
```

docs/aws.md
```diff
@@ -0,0 +1,67 @@
...
+Amazon has an extensive documentation on registering or tranferring your domain to route53 <https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/Welcome.html>.
```
qdm12, line 50:
```suggestion
Amazon has [an extensive documentation on registering or tranfering your domain to route53](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/Welcome.html).
```

docs/aws.md
```diff
@@ -0,0 +1,67 @@
...
+}
+```
+
+### Compulsory parameters
+
+- `"domain"`
+- `"host"` is your host and can be a subdomain or `"@"` or the wildcard `"*"`
+- `"aws_access_key_id"`
+- `"aws_secret_access_key"`
+- `"hosted_zone_id"`
```
qdm12, line 34: Could this be replaced with `zone_id` perhaps? 🤔

docs/aws.md
```diff
@@ -0,0 +1,67 @@
+# AWS
+
+## Configuration
+
+### Example
+
+<!-- UPDATE THIS JSON EXAMPLE -->
```
qdm12, line 7:
```suggestion
```

qdm12 (commented):
Thanks for the PR! It looks quite good, I just wonder if it could be achieved without using their Go SDK to rely on less dependencies (call me crazy, that's fair enough as well!)

internal/provider/providers/aws/provider.go
marcelohpf: No specific reason for that. I just got watherver region UI was showing at the moment for the 'Global Access', I will make it mandatory.

internal/provider/providers/aws/provider.go
marcelohpf: It was just an easier way to find the requests on CloudTrail. It was more on personal notes and I forgot to remove. I will remove it.

internal/provider/providers/aws/provider.go
marcelohpf: Yes, it is possible to use the Route53 API directly. Code will get a little bigger because of the signature generation thou.

I was not aware of that guideline, I will try to include it in the contributing document.

docs/aws.md
marcelohpf: I saw other comments about this as well.

My idea was to make it more familiar to AWS users, these names are well known by users and brainless when using a tool: "Oh... so this is the key and that is the secrets".

docs/aws.md
marcelohpf: I tried using AWS specific names, I can replace this by `zone_id` and update the doc to reference the HostedZone ID

internal/provider/providers/aws/provider.go
marcelohpf: I found out that signing requests for Global Resources must be `us-east-1`. As the goal is to `UPSERT`, the ddns-updater should make requests only to `us-east-1`. I will remove the Region from options.

internal/provider/errors/validation.go
```diff
@@ -20,13 +20,15 @@ var (
 	ErrHostNotSet             = errors.New("host is not set")
 	ErrHostOnlySubdomain      = errors.New("host can only be a subdomain")
 	ErrHostWildcard           = errors.New(`host cannot be a "*"`)
+	ErrHostedZoneIDNotSet     = errors.New("hosted zone id is not set")
```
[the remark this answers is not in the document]
marcelohpf: I updated as requested. Please, take a look

internal/provider/errors/validation.go
```diff
@@ -20,13 +20,15 @@ var (
...
 	ErrIPv4KeyNotSet          = errors.New("IPv4 key is not set")
 	ErrIPv6KeyNotSet          = errors.New("IPv6 key is not set")
 	ErrKeyNotSet              = errors.New("key is not set")
 	ErrKeyNotValid            = errors.New("key is not valid")
 	ErrNameNotSet             = errors.New("name is not set")
 	ErrPasswordNotSet         = errors.New("password is not set")
 	ErrPasswordNotValid       = errors.New("password is not valid")
+	ErrSecretAccessKeyNotSet  = errors.New("secret access key is not set")
```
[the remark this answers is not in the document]
marcelohpf: Updated to use existing errors, please take another look.

internal/provider/providers/aws/provider.go
marcelohpf: Updated the region to be `us-east-1` due to the limitations. Please, take another look

internal/provider/providers/aws/provider.go
marcelohpf: Sure thing. I thought aws would be easier to people find it, route53 works fine as well. Updated.

internal/provider/providers/aws/provider.go
marcelohpf: As requested, I removed the SDK dependency, code is a bit bigger now, no need for change anymore I suppose.

internal/provider/providers/aws/provider.go
marcelohpf: New implementation uses error messages from Amazon itself. Let me know your thoughts on this.

docs/aws.md
marcelohpf: Updated to `zone_id`. Please, take another look.

internal/provider/providers/aws/provider.go
qdm12: This is awesome! Congrats! 💯

internal/provider/providers/aws/provider.go
qdm12: It's not really in the guideline, I'll add it!

qdm12:
First of all, my apologies for the long delay!

Second, it was very enjoyable to review your quality Go code, congrats on that! A lot of attention to details and good work on porting code from the AWS sdk to this repo 💯 

- I rebased your branch on the master branch 
- I reworked your code to be a bit smaller/simpler. For example the headers sanitization got replaced by just hardcoded constants, which fit fine in this use case AFAIK.
- I added tests for the signer, although I am not 100% certain of the test expected values since I cannot really test it... And so, can you please test the new code and see if it still works fine (force an update by setting your record to 127.0.0.1 on the AWS UI first for example)?? Thanks!! - you can get the tip of the rebased branch by doing `git fetch && git reset --hard origin/feat/aws-simple-record` 😉 
- In the end, I decided to keep it as it is regarding the `aws_` prefix part. The reason is the environment variables do have the `AWS_` prefix to separate them from other programs, but in this case, it's already separated, so it doesn't make much sense. Anyway, that's just left as you left it last (without the prefix) 😉

marcelohpf:
> * which fit fine in this use case AFAIK

Yes, by their documentation, it wouldn't be a problem.

> And so, can you please test the new code

DNS was updated correctly. Everything is looking fine.

qdm12:
Perfect, thanks for the feedback, merging it 🚀 !

Issue #485, opened 2023-06-14
Feature request: Specify an update period per provider


qdm12:
Be able to specify an update period per provider which would override the global update period, if set.

This is due to some providers such as zoneedit.com which requires a minimum update period of 10 minutes, and some might be lower or higher.

PR #774, opened 2024-07-27
Provider Porkbun: Delete Default Parked DNS Entry for *.domain.tld

bentemple:
# Description:
Porkbun creates default ALIAS and CNAME domain records pointing to `pixie.porkbun.com` (Porkbun's parked domain website)

The current logic flow prior to this PR would look for an A or AAAA domain record, and if none exists, attempt to delete the ALIAS record for any subdomain and then proceed with the record creation.
This updates the logic flow to only look for a conflicting ALIAS record for the top level `domain.tld`, and a conflicting CNAME record for the `*.domain.tld`. Additionally, we verify that the content of this record matches `pixie.porkbun.com` and we only delete for the expected default value.
If the value does not match the expected `pixie.porkbun.com` we produce more helpful error messages.

Note: Any other domain name we simply attempt to create a record if no record exists, which will return a 400 error with no helpful error reason if there is a conflicting record of another type. We may want to consider checking for any conflicting records to produce a more helpful error for the user, but that change is well outside of the scope of this PR, and should likely be done in a more general way and apply to all DNS providers, rather than porkbun exclusively.

# Test-Plan:
Created a new domain.tld on Porkbun
Verified the default records were created:
`ALIAS domain.tld -> pixie.porkbun.com`
`CNAME *.domain.tld -> pixie.porkbun.com`
Started DDNS-Updater
Verified that both domain records were successfully deleted and updated

Reset the ALIAS domain record to point to `not-pixie.porkbun.com`
Reset the CNAME domain record to point to `not-pixie.porkbun.com`
Started DDNS-Updater
Verified that both domain records failed with the expected conflicting record error message.

screenshot_2024-08-17-0210 20


bentemple:
As commented elsewhere, probably should scrap this and just move the logic higher in the program. i.e. on a failed check, after 5 minutes re-check if the domain is in-fact wrong and needs updating still.

Bananas-Are-Yellow:
> Ultimately simply restarting the DDNS-Updater service after the domain has successfully propogated will resolve the issue

I'm having the 400 error problem, but the problem is not resolved by restarting the container with:
```
docker compose stop && docker compose up -d
```
The `updates.json` file shows the correct IP address recorded a few times, but the 400 error still occurs once the A records have been set. It's as if the `updates.json` file is not being read to see if the IP address has changed.

Here is my `config.json` file.

```
{
    "settings": [
        {
            "provider": "porkbun",
            "domain": "example.com",
            "host": "@,*",
            "api_key": "pk1_...",
            "secret_api_key": "sk1_...",
            "ip_version": "ipv4",
            "ipv6_suffix": ""
        }
    ]
}
```

qdm12:
> As commented elsewhere, probably should scrap this and just move the logic higher in the program. i.e. on a failed check, after 5 minutes re-check if the domain is in-fact wrong and needs updating still.

I think #485 would solve it, and we can log a warning if that period is less than the ttl of the record. I will mark it as 🚨 urgent, but will hold off a few days to start working on it (got other issues to attend to!)

> this would get into a scenario where ostensibly you might be failing to update 1 provider, but another provider has the correct DNS, and so it obfuscates the error.

One more reason we need a period per setting to clarify/simplify all this 😄 

@Bananas-Are-Yellow  First of all bananas can be brown and black too 🍌 !

> I'm having the 400 error problem

That's odd, @bentemple any idea why that could be? @Bananas-Are-Yellow can you share your logs, running the container with `LOG_LEVEL=debug` (⚠️ and hide your credentials)

> It's as if the `updates.json` file is not being read to see if the IP address has changed.

It's not, the program resolves the domain name (abc.domain.com) every period and compares it to your public ip address. updates.json is only used to store changes for the web ui AND is used by cloudflare when `proxied` is enabled, since ddns-updater cannot resolve the domain (since it's proxied).

Bananas-Are-Yellow:
It looks like I misunderstood how `updates.json` is used. I thought it was used as a record of what had already been set, to avoid setting the same value again. My confusion came from this:

> Ultimately simply restarting the DDNS-Updater service after the domain has successfully propogated will resolve the issue, but this can be quite confusing for someone who is dealing with setup issues, removes the updates.json and then gets into a bad state where the domain is constantly returning a 400 error and it never resolves itself without a service restart.

For me, "restarting the DDNS-Updater service after the domain has successfully propogated", makes no difference. But the main issue is certainly true:

> Porkbun will return a 400 error if you attempt to update the IP for a DNS record that is already set to the same value.

That's what I see too.

qdm12:
We may end up merging this, since there is also another person with the same problem on Porkbun. But I would ideally like to understand the root cause of this since it may apply to other registrars. I'm also wondering if this is specific to porkbun being slow at propagating updates 🤔 since I have not seen similar issues for other registrars.

bentemple:
So, when I was dealing with this, I also had a bunch of wonkiness because of the wildcard bug. It's actually why I made all these various fixes, but as I teased out what was happening it started to make more sense, and for example with this bug, I don't really think this _should_ be necessary, as it only applies to users who have removed their `updates.json` file before the timeout has expired. 

@Bananas-Are-Yellow Logs would be super helpful, also what version of DDNS-Updater are you using?

bentemple:
version shouldn't matter actually. I just retested on v2.7.0, and removing my updates.json file, then re-running correctly finds the dns as not needing to be updated and it does nothing.

bentemple:
I wonder if it actually makes sense to update the logic to:
```
if (noRecords)
  if ( top level domain ) 
      look for domain.tld ALIAS
      if ALIAS == pixie.porkbun.com
        delete
      else 
        logical error denoting the conflicting entry and user input needed
  if ( *.domain.tld ) 
    look for *.domain.tld CNAME
    if CNAME == pixie.porkbun.com
      delete
    else
      logical error denoting the conflicting CNAME entry and user input needed
```
Ultimately, we want to create the A / AAAA entries if none exist, but we really only want to be messing with CNAME and ALIAS conflicts if they're expected values. Just need to queue the user in better to the reason for the failure, and sadly the porkbun API generates worthless error messages. "400, something went wrong"

qdm12:
> I wonder if it actually makes sense to update the logic to

TLDR: let's do your version!

Longer story:
I've commented my own version in https://github.com/qdm12/ddns-updater/pull/776#issuecomment-2275194584 which is quite similar. I just don't delete unless there is a 400 code, so the API calls count is 2 (get records+create) to 4 (get records+create+deletion+create).
In your case, API calls count is 3 (get records+get alias|cname+create) to 4 (get records+get alias|cname+delete+create) but it adds the safety to check the alias|cname are the default porkbun ones. Plus creation of records are rare so API calls count doesnt' matter much really.

Bananas-Are-Yellow:
Yes, I removed the CNAME and ALIAS records that Porkbun provided when I registered my domain.

Yes, `nslookup` gives the correct IP address once it has been set.

Here's my test. I deleted `updates.json` to start from a clean sheet, and I deleted all my A records in the Porkbun website, so that I only have some MX and TXT records remaining. I have simplified my `config.json` to a single `bananasareyellow.com` domain (no prizes for guessing that this was my domain name), because I previously had a wildcard in there too.
```
{
    "settings": [
        {
            "provider": "porkbun",
            "domain": "bananasareyellow.com",
            "api_key": "pk1_...",
            "secret_api_key": "sk1_...",
            "ip_version": "ipv4",
            "ipv6_suffix": ""
        }
    ]
}
```
This successfully creates the A record. Then, 10 minutes later, it tries to set the same IP address and this fails with error 400.

I think I can see why. So that I can use my domain name from inside my network, I have a CNAME record set in my Pi-hole DNS so that `bananasareyellow.com` redirects to my home server machine where my docker containers are running. This seems to be confusing `ddns-updater` because it looks up the current IP address and finds the internal IP address of 192.168.1.8:
```
2024-08-08T08:13:18Z DEBUG configured to fetch IP: v4 or v6: false, v4: true, v6: false
2024-08-08T08:13:19Z DEBUG your public IP address are: v4 or v6: invalid IP, v4: 168.86.199.193, v6: invalid IP
2024-08-08T08:13:19Z INFO ipv4 address of bananasareyellow.com is 192.168.1.8 and your ipv4 address  is 168.86.199.193
2024-08-08T08:13:19Z INFO Updating record [domain: bananasareyellow.com | owner: @ | provider: porkbun | ip: ipv4] to use 168.86.199.193
```
I've tried setting `PUBLICIP_DNS_PROVIDERS=cloudflare`, thinking that this would bypass my Pi-hole DNS, but this makes no difference.

Here is the debug log (with `PUBLICIP_DNS_PROVIDERS=cloudflare` set). I've hidden the keys, of course.

[_porkbun-ddns-updater-1_logs.txt](https://github.com/user-attachments/files/16540284/_porkbun-ddns-updater-1_logs.txt)

Is there a way to prevent the docker container from using my Pi-hole DNS?

qdm12:
The `PUBLICIP*` options are to get your public IP address, not to resolve your domain. Try `RESOLVER_ADDRESS=1.1.1.1:53` so it goes around pihole.

Bananas-Are-Yellow:
To answer my own question, yes, there is a way to prevent `ddns-updater` from using my Pi-hole DNS. Just add a `dns` section to the docker compose file. Here I am using Cloudflare DNS servers `1.1.1.1` and `1.0.0.1`.
```
services:
  ddns-updater:
    image: qmcgaw/ddns-updater
    container_name: ddns-updater
    network_mode: bridge
    ports:
      - 8000:8000/tcp
    volumes:
      - ./data:/updater/data
    dns:
      - 1.1.1.1
      - 1.0.0.1
    environment:
      ...
```
Now it is working fine, as can be seen in the log:
```
2024-08-08T09:33:40Z DEBUG configured to fetch IP: v4 or v6: false, v4: true, v6: false
2024-08-08T09:33:40Z DEBUG your public IP address are: v4 or v6: invalid IP, v4: 168.86.199.193, v6: invalid IP
2024-08-08T09:33:40Z DEBUG ipv4 address of bananasareyellow.com is 168.86.199.193 and your ipv4 address is 168.86.199.193, skipping update
```
That fixes the problem.

> The `PUBLICIP*` options are to get your public IP address, not to resolve your domain. Try `RESOLVER_ADDRESS=1.1.1.1:53` so it goes around pihole.

I didn't try this alternative approach. Maybe that would fix it too, but I think a `dns` section in the docker compose file is clearer (to me at least).

qdm12:
> I didn't try this alternative approach. Maybe that would fix it too, but I think a dns section in the docker compose file is clearer (to me at least).

Eh it depends, I'd say the opposite, since the option only affects the http client resolving the domain names. For example resolving API domain names such as porkbun.com uses the normal dns. In my case I run dns over tls and prefer to use that for API calls for example. Anyway, marking your and my comments as off topic since it's resolved/a bit out topic.

bentemple:
> > I wonder if it actually makes sense to update the logic to
> 
> TLDR: let's do your version!
> 
> Longer story: I've commented my own version in [#776 (comment)](https://github.com/qdm12/ddns-updater/pull/776#issuecomment-2275194584) which is quite similar. I just don't delete unless there is a 400 code, so the API calls count is 2 (get records+create) to 4 (get records+create+deletion+create). In your case, API calls count is 3 (get records+get alias|cname+create) to 4 (get records+get alias|cname+delete+create) but it adds the safety to check the alias|cname are the default porkbun ones. Plus creation of records are rare so API calls count doesnt' matter much really.

sounds good. I'll update it accordingly. Sorry went to Defcon this last week so just busy life.

bentemple:
In regards to @Bananas-Are-Yellow 's issue: 
image

qdm12:
> sounds good. I'll update it accordingly. Sorry went to Defcon this last week so just busy life.

No pressure at all, the existing version already works more or less fine 😉

Ps: I saved that picture 😄

internal/provider/providers/porkbun/provider.go
```diff
@@ -138,8 +153,14 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
 		return ip, nil
 	}
 
-	for _, recordID := range recordIDs {
-		err = p.updateRecord(ctx, client, recordType, ipStr, recordID)
+	for _, record := range records {
+		if record.Content == ipStr {
```
bentemple, line 157: revert

internal/provider/providers/porkbun/provider.go
```diff
@@ -148,17 +163,31 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
 	return ip, nil
 }
 
-func (p *Provider) deleteALIASRecordIfNeeded(ctx context.Context, client *http.Client) (err error) {
-	aliasRecordIDs, err := p.getRecordIDs(ctx, client, "ALIAS")
+// Deletes the matching record for a specific recordType iff the content matches the expected content value.
+// Returns an error if there is a matching record but an unexpected content value.
+// If there is no matching record, there is nothing to do.
```
qdm12, line 168: nit comment a bit improved.

```suggestion
// deleteMatchingRecord deletes an eventually present record matching a specific record type if the content matches the expected content value.
// It returns an error if multiple records are found or if one record is found with an unexpected value.
```

internal/provider/providers/porkbun/api.go
```diff
@@ -12,9 +12,20 @@ import (
...
+	Content string `json:"content"`
+	// TTL      string `json:"ttl"`
+	// Priority string `json:"prio"`
+	// Notes    string `json:"notes"`
+}
+
 // See https://porkbun.com/api/json/v3/documentation#DNS%20Retrieve%20Records%20by%20Domain,%20Subdomain%20and%20Type
-func (p *Provider) getRecordIDs(ctx context.Context, client *http.Client, recordType string) (
-	recordIDs []string, err error) {
+func (p *Provider) getRecords(ctx context.Context, client *http.Client, recordType string) (
```
qdm12, line 27: Let's wrap the recordType in each error returned by `getRecords`, and remove it from wrappings from callers of this function. Since it's this function handling recordType, it should be the only one wrapping errors with it (this helps a lot as a general rule to write non-repetitive error wrappings).

Also my bad this was not case before!

internal/provider/providers/porkbun/provider.go
```diff
@@ -119,16 +119,31 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
...
 	if err != nil {
 		return netip.Addr{}, fmt.Errorf("getting record IDs: %w", err)
 	}
 
-	if len(recordIDs) == 0 {
-		// ALIAS record needs to be deleted to allow creating an A record.
-		err = p.deleteALIASRecordIfNeeded(ctx, client)
-		if err != nil {
-			return netip.Addr{}, fmt.Errorf("deleting ALIAS record if needed: %w", err)
+	if len(records) == 0 {
```
qdm12, line 127: nit perhaps let's flip this around, especially because the ` len(records) == 0` block is getting pretty big so have

```go
if len(records) > 0 {
	for _, record := range records {
		err = p.updateRecord(ctx, client, recordType, ipStr, recordID)
		if err != nil {
			return netip.Addr{}, fmt.Errorf("updating record: %w", err)
		}
	}
	return ip, nil
}

// For new domains, Porkbun Creates 2 Default DNS Records which point to their "parked" domain page:
// ...
```

internal/provider/providers/porkbun/provider.go
```diff
@@ -148,17 +163,31 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
...
-	aliasRecordIDs, err := p.getRecordIDs(ctx, client, "ALIAS")
+// Deletes the matching record for a specific recordType iff the content matches the expected content value.
+// Returns an error if there is a matching record but an unexpected content value.
+// If there is no matching record, there is nothing to do.
+func (p *Provider) deleteMatchingRecord(
+	ctx context.Context,
+	client *http.Client,
+	recordType string,
+	expectedContent string,
+) error {
```
qdm12, line 174: nit are you using a specific formatter for this 🤔 

Is it ok if we leave it as the existing formatting style

```suggestion
func (p *Provider) deleteMatchingRecord(ctx context.Context, client *http.Client, 
    recordType, expectedContent string) (err error) {
```

internal/provider/providers/porkbun/provider.go
```diff
@@ -148,17 +163,31 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
...
+	records, err := p.getRecords(ctx, client, recordType)
 	if err != nil {
-		return fmt.Errorf("getting ALIAS record IDs: %w", err)
-	} else if len(aliasRecordIDs) == 0 {
-		return nil
+		return fmt.Errorf("getting %s records: %w", recordType, err)
```
qdm12, line 177: see above comment - getRecords should wrap the record type

```suggestion
		return fmt.Errorf("getting records: %w", err)
```

internal/provider/providers/porkbun/provider.go
```diff
@@ -148,17 +163,31 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
 	return ip, nil
 }
 
-func (p *Provider) deleteALIASRecordIfNeeded(ctx context.Context, client *http.Client) (err error) {
-	aliasRecordIDs, err := p.getRecordIDs(ctx, client, "ALIAS")
+// Deletes the matching record for a specific recordType iff the content matches the expected content value.
+// Returns an error if there is a matching record but an unexpected content value.
+// If there is no matching record, there is nothing to do.
+func (p *Provider) deleteMatchingRecord(
```
qdm12, line 169: to make more sense, this function should be named `deleteSingleMatchingRecord`, since it doesn't delete all records matching an expected value, and is meant to only handle a single record.

internal/provider/providers/porkbun/provider.go
```diff
@@ -148,17 +163,31 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
...
+			return fmt.Errorf("deleting %s record: %w", recordType, err)
+		}
+		return nil
+	} else if len(records) >= 1 {
+		// We have 1 or more matching records, but the expectedContent didn't match. Throw a conflicting record error.
+		return fmt.Errorf("deleting %s record: %w", recordType, errors.ErrConflictingRecord)
 	}
+
+	// No record found for recordType, nothing to do.
 	return nil
```
qdm12, line 192: Changed to a switch block for better clarity I hope + changed error wrappings a bit

```suggestion
	switch {
	case len(records) == 0:
		return nil
	case len(records) > 1:
		return fmt.Errorf("%w: %d %s records are already set", errors.ErrConflictingRecord, recordType)
	case records[0].Content != expectedContent:
		return fmt.Errorf("%w: %s record has content %q mismatching expected content %q",
			errors.ErrConflictingRecord, recordType, records[0].Content, expectedContent)
	}
	
	// Single record with content matching expected content.
	err = p.deleteRecord(ctx, client, recordType)
	if err != nil {
		return fmt.Errorf("deleting record: %w", err)
	}
	return nil
```

internal/provider/providers/porkbun/api.go
```diff
@@ -176,12 +181,12 @@ func (p *Provider) updateRecord(ctx context.Context, client *http.Client,
 }
 
 // See https://porkbun.com/api/json/v3/documentation#DNS%20Delete%20Records%20by%20Domain,%20Subdomain%20and%20Type
-func (p *Provider) deleteAliasRecord(ctx context.Context, client *http.Client) (err error) {
+func (p *Provider) deleteRecord(ctx context.Context, client *http.Client, recordType string) (err error) {
```
qdm12, line 184: Same as createRecord, the error returned should contain the recordType as wrapping 🙏

internal/provider/providers/porkbun/provider.go
```diff
@@ -119,16 +119,31 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
...
+			// Delete ALIAS domain.tld -> pixie.porkbun.com record
+			err = p.deleteMatchingRecord(ctx, client, constants.ALIAS, porkbunParkedDomain)
+			if err != nil {
+				return netip.Addr{}, err
+			}
+		case p.owner == "*":
+			// Delete CNAME *.domain.tld -> pixie.porkbun.com record
+			err = p.deleteMatchingRecord(ctx, client, constants.CNAME, porkbunParkedDomain)
+			if err != nil {
+				return netip.Addr{}, err
```
qdm12, line 145:
```suggestion
				return netip.Addr{}, fmt.Errorf("deleting default parked domain record: %w", err)
```

internal/provider/providers/porkbun/provider.go
```diff
@@ -119,16 +119,31 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
...
+		switch {
+		case p.owner == "@":
+			// Delete ALIAS domain.tld -> pixie.porkbun.com record
+			err = p.deleteMatchingRecord(ctx, client, constants.ALIAS, porkbunParkedDomain)
+			if err != nil {
+				return netip.Addr{}, err
+			}
+		case p.owner == "*":
+			// Delete CNAME *.domain.tld -> pixie.porkbun.com record
+			err = p.deleteMatchingRecord(ctx, client, constants.CNAME, porkbunParkedDomain)
```
qdm12, line 143: We should make sure this deletes only for the `*` owner

internal/provider/providers/porkbun/provider.go
```diff
@@ -119,16 +119,31 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
...
+		// For new domains, Porkbun Creates 2 Default DNS Records which point to their "parked" domain page:
+		// ALIAS domain.tld -> pixie.porkbun.com
+		// CNAME *.domain.tld -> pixie.porkbun.com
+		// ALIAS and CNAME records conflict with A and AAAA records, and attempting to create an A or AAAA record
+		// will return a 400 error if they aren't first deleted.
+		porkbunParkedDomain := "pixie.porkbun.com"
+		switch {
+		case p.owner == "@":
+			// Delete ALIAS domain.tld -> pixie.porkbun.com record
+			err = p.deleteMatchingRecord(ctx, client, constants.ALIAS, porkbunParkedDomain)
```
qdm12, line 137: We should make sure this deletes only for the `@` owner

qdm12 (changes requested):
Thanks!

A few comments, many nits (apologies) and a few important changes to make (without the `nit` prefix 😄). Feel free to skip `nit`s if you want.

internal/provider/providers/porkbun/provider.go
bentemple: I love early returns 💯

internal/provider/providers/porkbun/api.go
bentemple: I really like this philosophy. I'm hoping what I changed it to makes sense. In general I don't think I have good error message patterns.

Seems like as a general rule of thumb, the caller will error message with what it was invoking, i..e. update, create, and the function should return any relevant error values that will form the context of the error. So if you pass a value to the function, it's more or less expected that the error returned will contain that relevant info, whereas the error log of the caller will include the context of the function itself to avoid duplicating that across every error log inside the function?

i.e. obviously we could put the context of updating, deleting, and creating as well in each error log, and simply log the error in its entirety in the parent without any additional message changes, but then  you get all those strings duplicated in every message.

Does that sound about right?

internal/provider/providers/porkbun/api.go
bentemple: 👍

internal/provider/providers/porkbun/provider.go
```diff
@@ -119,16 +119,31 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
...
+			if err != nil {
+				return netip.Addr{}, err
+			}
+		case p.owner == "*":
+			// Delete CNAME *.domain.tld -> pixie.porkbun.com record
```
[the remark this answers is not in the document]
bentemple: Hmmmm, so I do agree, but it shouldn't error out if there IS an alias that's set to something else... right?

Additionally, as it's currently coded we use the `p.owner` to get records, so will probably need to start passing in what domain we want to fetch / delete for instead of using the current `p.owner` does that make sense / seem like the direction you want to go? would certainly make the API less tied to the context of the provider.

internal/provider/providers/porkbun/provider.go
bentemple: Are you thinking we should pass in the owner we want to delete for? p.deleteMatchingRecord uses p.getRecords which uses p.owner, so as it's coded currently, this could only be for the  `@` owner

internal/provider/providers/porkbun/provider.go
bentemple: same as above, regarding using the `p.owner` throughout all the API methods, happy to update to pass in what domain to delete for.

bentemple:
I took a pass at deleting the `@` ALIAS opportunistically when setting the '*' domain. From what I could tell, this required updating the API to pass in the owner, but maybe there's a better way I just wasn't seeing.

I kept that separate as it's own [commit](https://github.com/qdm12/ddns-updater/pull/774/commits/d286bcdc8ab3a005d25ffeea1d523a5797fbaf27), so it's both easy to see the change, and we can trivially revert it if it's not the direction you were thinking. 


Side-note: I really appreciate all the nits. Code homogeneity is really important, and nits are a really great way to level up code quality (imo)

internal/provider/providers/porkbun/api.go
```diff
@@ -32,54 +42,48 @@ func (p *Provider) getRecordIDs(ctx context.Context, client *http.Client, record
...
 	response, err := client.Do(request)
 	if err != nil {
-		return nil, fmt.Errorf("doing http request: %w", err)
+		return nil, fmt.Errorf("doing %s http request: %w", recordType, err)
 	}
 	defer response.Body.Close()
 
 	if response.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("%w: %d: %s", errors.ErrHTTPStatusNotValid,
+		return nil, fmt.Errorf("%s - %w: %d: %s", recordType, errors.ErrHTTPStatusNotValid,
```
bentemple, line 68: Wasn't entirely sure the best way to format these error messages to include the recordType. I welcome any input

internal/provider/providers/porkbun/provider.go
```diff
@@ -119,46 +119,76 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
...
-		// ALIAS record needs to be deleted to allow creating an A record.
-		err = p.deleteALIASRecordIfNeeded(ctx, client)
-		if err != nil {
-			return netip.Addr{}, fmt.Errorf("deleting ALIAS record if needed: %w", err)
-		}
-
-		err = p.createRecord(ctx, client, recordType, ipStr)
-		if err != nil {
-			return netip.Addr{}, fmt.Errorf("creating record: %w", err)
+	if len(records) > 0 {
```
qdm12: Reverted to `len(recordIDs) == 0` given the new function `deleteDefaultConflictingRecordsIfNeeded` not taking as many lines in the if block 😉 (to have less deltas)

internal/provider/providers/porkbun/provider.go
```diff
@@ -119,46 +119,76 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
...
 		return ip, nil
 	}
 
-	for _, recordID := range recordIDs {
-		err = p.updateRecord(ctx, client, recordType, ipStr, recordID)
+	// For new domains, Porkbun Creates 2 Default DNS Records which point to their "parked" domain page:
+	// ALIAS domain.tld -> pixie.porkbun.com
+	// CNAME *.domain.tld -> pixie.porkbun.com
+	// ALIAS and CNAME records conflict with A and AAAA records, and attempting to create an A or AAAA record
+	// will return a 400 error if they aren't first deleted.
```
qdm12: Resumed the comment with

> // deleteDefaultConflictingRecordsIfNeeded deletes any default records that would conflict with a new record,
> // see https://github.com/qdm12/ddns-updater/blob/master/docs/porkbun.md#record-creation

internal/provider/providers/porkbun/provider.go
```diff
@@ -119,46 +119,76 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
...
+	porkbunParkedDomain := "pixie.porkbun.com"
```
qdm12: changed this to

```go
const porkbunParkedDomain = "pixie.porkbun.com"
```

because constants are always nicer if possible 😉

internal/provider/providers/porkbun/provider.go
```diff
@@ -119,46 +119,76 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
...
+	switch {
```
qdm12: changed to `switch p.owner {` directly

internal/provider/providers/porkbun/provider.go
```diff
@@ -119,46 +119,76 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
...
+	case p.owner == "@":
+		// Delete ALIAS domain.tld -> pixie.porkbun.com record
+		err = p.deleteSingleMatchingRecord(ctx, client, constants.ALIAS, "@", porkbunParkedDomain)
+		if err != nil {
+			return netip.Addr{}, fmt.Errorf("deleting default parked domain record: %w", err)
```
qdm12: changed error wrapping to

```go
fmt.Errorf("deleting default ALIAS @ parked domain record: %w", err)
```

internal/provider/providers/porkbun/provider.go
```diff
@@ -119,46 +119,76 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
...
+		}
+	case p.owner == "*":
+		// Delete ALIAS domain.tld -> pixie.porkbun.com record
+		// Error is ignored as the ALIAS could be set to something besides the parked domain. Failure here is non-fatal.
+		_ = p.deleteSingleMatchingRecord(ctx, client, constants.ALIAS, "@", porkbunParkedDomain)
```
qdm12: Eh it should be a fatal error if we get an API error for example, so I changed it to

```go
		err = p.deleteSingleMatchingRecord(ctx, client, constants.ALIAS, "@", porkbunParkedDomain)
		if err != nil && !stderrors.Is(err, errors.ErrConflictingRecord) {
			// allow conflict ALIAS records to be set to something besides the parked domain
			return fmt.Errorf("deleting default ALIAS @ parked domain record: %w", err)
		}
```

qdm12 (commented):
Thanks for the re-work! And sorry for the 3 weeks delay, but I'm more or less back on the keyboard now. 😉

qdm12 (approved):
I allowed myself to push that generic http post function refactor commit to the master branch at 8c1b3e556cf6fb05471e332a2bb12e184aa0600d in order to only have relevant deltas in this PR, and then rebased your branch on the master branch resolving conflicts. I also pushed 2 small-ish commits on top of your commits.

Please review the code and perhaps test the program to double check it works fine (even basic update should do the test trick), and then I'm happy to merge it 😉

bentemple:
> I allowed myself to push that generic http post function refactor commit to the master branch at [8c1b3e5](https://github.com/qdm12/ddns-updater/commit/8c1b3e556cf6fb05471e332a2bb12e184aa0600d) in order to only have relevant deltas in this PR, and then rebased your branch on the master branch resolving conflicts. I also pushed 2 small-ish commits on top of your commits.
> 
> Please review the code and perhaps test the program to double check it works fine (even basic update should do the test trick), and then I'm happy to merge it 😉

no problem. I'll re-run my testplan tonight and push it up

qdm12:
All good then? Thanks! 💯

bentemple:
So sorry for the long delay. life has been crazy.

qdm12:
Hey no problem 🎉 !

One tiny nitpicky thing for next PR: merge with a ['git commit message'](https://www.conventionalcommits.org/en/v1.0.0/#summary), for example it would had been `fix(porkbun): delete default parked DNS entries if needed` (since it's not only for the wildcard as well), and resume final changes as bullet points instead of using the PR description. For example:

- delete default `@` ALIAS entry when creating a record for the root domain or the wildcard domain
- delete default `*` CNAME entry when creating a record for the wildcard domain

It's fine though, I've already copied/pasted this for the next draft of the next release 😉

bentemple:
> Hey no problem 🎉 !
> 
> One tiny nitpicky thing for next PR: merge with a ['conventional commit message'](https://www.conventionalcommits.org/en/v1.0.0/#summary), for example it would had been `fix(porkbun): delete default parked DNS entries if needed` (since it's not only for the wildcard as well), and resume final changes as bullet points instead of using the PR description. For example:
> 
>     * delete default `@` ALIAS entry when creating a record for the root domain or the wildcard domain
> 
>     * delete default `*` CNAME entry when creating a record for the wildcard domain
> 
> 
> It's fine though, I've already copied/pasted this for the next draft of the next release 😉

Yeah... :( I looked at all of the other commit messages after merging and realized that I hadn't followed the convention. I'll def follow the convention in the future. Thanks for pointing out :)

PR #775, opened 2024-07-27
Provider Porkbun: Remove trailing '.' from alias delete request

bentemple:
# Description
The porkbun documentation example request to delete a domain does not contain a trailing '.' `URI Endpoint Example: https://api.porkbun.com/api/json/v3/dns/deleteByNameType/borseth.ink/A/www`

This updates the code to match that.

# Test-Plan
Changed delete to allow arbitrary deletes of any type (instead of just ALIAS)
Changed code to delete+create instead of update to change a porkbun domain Verified that delete works by deleting and creating a domain to update it instead of calling the update endpoint.


qdm12 (approved):
Thanks for splitting your PR 💯 👍

PR #776, opened 2024-07-27
Provider Porkbun: Update delete API to allow arbitrary deletions

bentemple:
# Description
This updates the API to allow calling delete on any type. It's currently unused, but more closely matches how the API is documented.

# Test-Plan
Used delete to delete A records and Alias records

### DIFF
--- a/internal/provider/providers/porkbun/api.go
+++ b/internal/provider/providers/porkbun/api.go
@@ -172,7 +172,7 @@ func (p *Provider) updateRecord(ctx context.Context, client *http.Client,
 }
 
 // See https://porkbun.com/api/json/v3/documentation#DNS%20Delete%20Records%20by%20Domain,%20Subdomain%20and%20Type
-func (p *Provider) deleteAliasRecord(ctx context.Context, client *http.Client) (err error) {
+func (p *Provider) deleteRecord(ctx context.Context, client *http.Client, recordType string) (err error) {
 	var subdomain string
 	if p.owner != "@" {
 		subdomain = p.owner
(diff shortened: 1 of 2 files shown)


bentemple:
I understand that the decision to lock it down to just "ALIAS" may have been fully intentional, so feel free to just close the PR

qdm12:
Let's close this PR since I don't think it's really necessary (forgive me for being a simplicity maniac! 😄)

I also invited you as a collaborator to the project, since I ❤️ your code style, PR splitting, and comments 😉
No pressure to be active or anything, the project is rather healthy and issues aren't piling up quickly either, so it should be more fun to monitor than a burden 🤣

bentemple:
> Let's close this PR since I don't think it's really necessary (forgive me for being a simplicity maniac! 😄)
> 
> I also invited you as a collaborator to the project, since I ❤️ your code style, PR splitting, and comments 😉 No pressure to be active or anything, the project is rather healthy and issues aren't piling up quickly either, so it should be more fun to monitor than a burden 🤣

Thank you! I'll try to help out where I can, probably won't be very active, but I'll certainly PR everything exactly like I have here.

>forgive me for being a simplicity maniac!
Consistency is super critical, and I fully support having a methodology to everything

bentemple:
Realized the other day when I added a new domain that there is a second record that porkbun creates by default.
Porkbun creates the following records for new domains:

Screenshot 2024-08-07 at 09 47 36

I.e. using just the default porkbun setup, if I add `domain.tld` pointing to my IP, and we do nothing else, `test.domain.tld` will point to `pixie.porkbun.com`

Screenshot 2024-08-07 at 11 33 41

I don't actually think deleting it by default may be the best option. It's really a DNS mis-configuration (trying to set A record with a previously set CNAME)
I'm curious to hear if you have any thoughts on what the best course of action is to take here. My gut says, would be great to just have better error handling, and if we get a 400 error, check to see if there are CNAME or ALIAS records present and show a better error message in the UI or something.
At the very least we could just add a comment in the docs about these default records and to delete them if desired.

qdm12:
> I don't actually think deleting it by default may be the best option. It's really a DNS mis-configuration (trying to set A record with a previously set CNAME)

I'm not 100% sure I follow. We should however definitely document it and log it out when deleting, since we do destructive things. In my view, it would be nice to:

1. Try updating without considering these 2 pre-created records. If it succeeds all good (should be the case after the 1st run)
2. If we get a 400 code (by the way, do we get any interesting body error message in the response?):
    - The owner is `@`, then delete the alias record (alias only applies to the `@` owner right?), and re-attempt setting the record
    - The owner is `*`, then delete the cname record, and re-attempt setting the record

And in that case, we need this PR 😄 !

Issue #786, opened 2024-08-09
Documentation: Cloudflare configuration documentation requires some improvements.


abpai94:
**TLDR**: *Describe your issue in a one liner here*

1. Is this urgent: No
2. DNS provider(s) you use: Cloudflare
3. Program version:
`Running version 2.7.0 built on 2024-06-26T19:34:57.277Z (commit d37f057)`
4. What are you using to run the container: docker-compose

The Cloudflare documentation defined here: [Cloudflare README](https://github.com/qdm12/ddns-updater/blob/master/docs/cloudflare.md) was helpful to start configuring DDNS updates as follows:
```
{
    "settings": [
        {
            "provider": "cloudflare",
            "zone_identifier": "<zone_id>",
            "ttl": 600,
            "ip_version": "ipv4",
            "domain": "<domain>",
            "token": "<token>"
        },
        {
            "provider": "cloudflare",
            "zone_identifier": "<zone_id>",
            "ttl": 600,
            "ip_version": "ipv4",
            "domain": "<second_domain>",
            "token": "<token>"
        }
    ]
}
``` 
This produces the following log output:
```
2024-08-09T09:12:43+02:00 WARN cannot DNS resolve <domain> after 5 tries: lookup <domain>: no such host
2024-08-09T09:12:43+02:00 INFO ipv4 address of <domain> is invalid IP and your ipv4 address  is <IP>
2024-08-09T09:12:43+02:00 WARN cannot DNS resolve <second_domain after 5 tries: lookup <second_domain>: no such host
2024-08-09T09:12:43+02:00 INFO ipv4 address of <second_domain> is invalid IP and your ipv4 address  is <IP>
2024-08-09T09:12:43+02:00 INFO Updating record [domain: <domain | host:  | provider: cloudflare | ip: ipv4] to use <IP> 2024-08-09T09:12:44+02:00 ERROR creating record: unsuccessful result: error 1004: DNS Validation Error; 
2024-08-09T09:12:44+02:00 INFO Updating record [domain: <second_domain> | host:  | provider: cloudflare | ip: ipv4] to use <IP>
2024-08-09T09:12:45+02:00 ERROR creating record: unsuccessful result: error 1004: DNS Validation Error; 
2024-08-09T09:14:20+02:00 WARN Caught OS signal, shutting down
```

The issue appears to be that the DNS cannot be validated. The API key contains the require Zone.Zone and Zone.DNS to update the IP address for the domain.

[...]

PR #788, opened 2024-08-09
#786: Updated documentation for cloudflare to include host configuration for root and subdomains.

abpai94:
#786


qdm12:
See https://github.com/qdm12/ddns-updater/issues/786#issuecomment-2294834171

PR #803, opened 2024-09-05
The "host" key is needed

GeiserX:
I recently tried following these docs but I had to add "host" to make it work properly, as the updater was complaining about it.


qdm12:
You followed the wrong version for the documentation 😉 the latest image/program does not require that host key anymore. The docs is versioned, see the readme's top.

I appreciate the effort though, and also, I'm going to create a new release since this is not the first time this happens, to reduce the confusion with the host key requirement. If I may ask, how were you using the program? Standalone binary or docker image?

PR #810, opened 2024-09-09
Support for Domeneshop

jobrajac:
Added support for Domeneshop, a popular Norwegian DNS provider.
Tested and works:)

https://domene.shop/
https://api.domeneshop.no/docs/

### DIFF
--- a/internal/provider/providers/domeneshop/provider.go
+++ b/internal/provider/providers/domeneshop/provider.go
@@ -8,7 +8,6 @@ import (
 	"net/http"
 	"net/netip"
 	"net/url"
-	"strings"
 
 	"github.com/qdm12/ddns-updater/internal/models"
 	"github.com/qdm12/ddns-updater/internal/provider/constants"
@@ -20,7 +19,7 @@ import (
 
 type Provider struct {
 	domain     string
-	owner  string
+	owner      string
 	token      string
 	secret     string
 	ipVersion  ipversion.IPVersion
@@ -148,6 +147,7 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
 
 	switch response.StatusCode {
 	case http.StatusNoContent:
+		return ip, nil
 	case http.StatusNotFound:
 		return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrHostnameNotExists, utils.ToSingleLine(s))
 	default:
@@ -155,12 +155,4 @@ func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Add
 			errors.ErrHTTPStatusNotValid, response.StatusCode, utils.ToSingleLine(s))
 	}
 
-	switch {
-	case strings.HasPrefix(s, "Successful operation"):
-		return ip, nil
-	case strings.HasPrefix(s, "Domain not found"):
-		return netip.Addr{}, fmt.Errorf("%w", errors.ErrDomainNotFound)
-	default:
-		return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrUnknownResponse, s)
-	}
 }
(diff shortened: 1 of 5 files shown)


internal/provider/providers/domeneshop/provider.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	defer response.Body.Close()
+
+	b, err := io.ReadAll(response.Body)
+	if err != nil {
+		return netip.Addr{}, fmt.Errorf("reading response body: %w", err)
+	}
+	s := string(b)
+
+	defer response.Body.Close()
```
qdm12, line 146: Remove this it's already defered above

```suggestion
```

internal/provider/providers/domeneshop/provider.go
```diff
@@ -0,0 +1,158 @@
...
+	switch response.StatusCode {
+	case http.StatusNoContent:
+		return ip, nil
+	case http.StatusNotFound:
+		return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrHostnameNotExists, utils.ToSingleLine(s))
+	default:
+		return netip.Addr{}, fmt.Errorf("%w: %d: %s",
+			errors.ErrHTTPStatusNotValid, response.StatusCode, utils.ToSingleLine(s))
+	}
+
```
qdm12, line 157: Nit remove empty trail line

```suggestion
```

qdm12 (commented):
Nice thanks!

1. Is there an API Route to create records if they don't exist?
2. Does this code work with ipv6 (at least according to their documentation)
3. Wildcards are not supported correct?

Thanks!

jobrajac:
> Nice thanks!
> 
>     1. Is there an API Route to create records if they don't exist?
> 
>     2. Does this code work with ipv6 (at least according to their documentation)
> 
>     3. Wildcards are not supported correct?
> 
> 
> Thanks!

1. Yes, there is, but it is not something I have tested or have much use for personally. Is that a requirement for you to merge the branch or can that perhaps stay on the todo-list for this provider? As far I can see that option is not present for most of the other providers.

2. Yes. According to their docs you can set ipv4 or ipv6 using the same url param 'myip'.

3. I could not find any information on wildcards in their docs so I assumed it to be not supported. I did not test it though.

Thank you for the great repo and quick repsponse!

internal/provider/providers/domeneshop/provider.go
```diff
@@ -0,0 +1,158 @@
...
+	return p.owner
+}
+
+func (p *Provider) IPVersion() ipversion.IPVersion {
+	return p.ipVersion
+}
+
+func (p *Provider) IPv6Suffix() netip.Prefix {
+	return netip.Prefix{}
+}
```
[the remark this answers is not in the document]
jobrajac: Good catch, you're right. Fixed

jobrajac:
>     3. Can you try on your end with a wildcard record (assuming you're not needing 100% uptime/production environment)? Even if not possible for you to test it out, let's assume it's supported and wait for someone to create an issue about it if it doesn't work, instead of restricting it to no wildcard if we're not sure.

I tested with a wildcard update and it did not work. I can set a wildcard domain in the provider's gui, but the update endpoint will only return 404 unfortunately.

qdm12:
Thanks for checking the wildcard record! 👍
How about

> If the domain record doesn't exist, what error do they return?

I can jump in on your branch and implement that, no problem 😉 thanks!

jobrajac:
Hello. Sorry for the late reply. 

I checked the documentation again and it says the record is created if it does not exist. I don't know how I missed this the first time.

I tested it and it works fine:)

qdm12 (approved):

qdm12:
(I rebased your branch on the repository master branch, and pushed two tiny fixes to fix the CI)

- about to merge that PR, thanks! 💯

PR #829, opened 2024-10-02
feat(provider): add support for Vultr

amroessam:
First time writing go, this was easy thanks to great docs and example 🙏🏽 
Go easy on the roast 😛 

[Vultr DNS API docs for ref](https://www.vultr.com/api/#tag/dns/operation/create-dns-domain-record)

Excited for the merge, so I can use this instead of my janky script https://github.com/amroessam/vultr-ddns


amroessam:
Preemptively answering [these](https://github.com/qdm12/ddns-updater/pull/810#pullrequestreview-2292060445) questions

1. Yes, the default behavior is to create the record if it doesn't exist
2. Yes, I've tested it
3. Wildcards are supported but only for 1 record type, if it errors out it should reflect the error

ehab-qadah-lean (approved):

amroessam:
@qdm12 any luck getting to this? 🙏🏽

.gitignore
```diff
@@ -1 +1,4 @@
-/data
\ No newline at end of file
+/data
+.vscode
+app
+cmd/ddns-updater/data/
```
qdm12, line 4: Revert this 🙏  (or do another PR if you want to add i.e. `.vscode` to the .gitignore)

docs/vultr.md
```diff
@@ -0,0 +1,31 @@
+# Variomedia
```
qdm12, line 1: The typical mistake I do 50% of the time 😄 

```suggestion
# Vultr
```

docs/vultr.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Example
+
+```json
+{
+  "settings": [
+    {
+      "provider": "vultr",
+      "domain": "example.com",
+      "owner": "potato",
```
qdm12, line 13: The last program works with `domain` only, the owner (here `potato`) is automatically extracted from the domain string now.

```suggestion
      "domain": "potato.example.com",
```

docs/vultr.md
```diff
@@ -0,0 +1,31 @@
...
+### Compulsory parameters
+
+- `"domain"` is the domain to update. It can be `example.com` (root domain) or `potato.example.com` (subdomain of `example.com`).
+- `"token"` is your API key, can be obtained from the [account settings](https://my.vultr.com/settings/#settingsapi), this is used as a bearer token to authenticate your request
+
+### Optional parameters
+
+- `"ip_version"` can be `ipv4` (A records), or `ipv6` (AAAA records) or `ipv4 or ipv6` (update one of the two, depending on the public ip found). It defaults to `ipv4 or ipv6`.
+- `"ipv6_suffix"` is the IPv6 interface identifier suffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw public IPv6 address obtained is used in the record updating.
+- `"ttl"` default is 900
```
qdm12, line 31: Are you sure it's 900?

docs/vultr.md
```diff
@@ -0,0 +1,31 @@
...
+      "ttl": 300,
+      "ip_version": "ipv4"
+    }
+  ]
+}
+```
+
+### Compulsory parameters
+
+- `"domain"` is the domain to update. It can be `example.com` (root domain) or `potato.example.com` (subdomain of `example.com`).
```
qdm12, line 24: Should the wildcard be documented as well, indicating the limitation on record type?

internal/provider/providers/vultr/provider.go
```diff
@@ -0,0 +1,195 @@
...
+	ipVersion  ipversion.IPVersion
+	ipv6Suffix netip.Prefix
+	token      string
+	ttl        uint32
+}
+
+func New(data json.RawMessage, domain, owner string,
+	ipVersion ipversion.IPVersion, ipv6Suffix netip.Prefix) (
+	provider *Provider, err error) {
+
```
qdm12, line 32: nit no need for first empty line

```suggestion
```

internal/provider/providers/vultr/provider.go
```diff
@@ -0,0 +1,195 @@
...
+func validateSettings(domain, owner, token string, ttl uint32) (err error) {
+	err = utils.CheckDomain(domain)
+	if err != nil {
+		return fmt.Errorf("%w: %w", errors.ErrDomainNotValid, err)
+	}
+
+	switch {
+	case owner == "":
+		return fmt.Errorf("%w", errors.ErrOwnerNotSet)
```
qdm12, line 65: nit I believe this can be removed, since it defaults to `@` if it's not set (in code called before creating this provider). Feel free to double check though.


```suggestion
```

internal/provider/providers/vultr/provider.go
```diff
@@ -0,0 +1,195 @@
...
+	case token == "":
+		return fmt.Errorf("%w", errors.ErrAPIKeyNotSet)
```
qdm12, line 67: Isn't there an `ErrTokenNotSet`? If not, please create one 🙏 Otherwise it's rather confusing for the user, since it's a `token` and the error talks about an api key.

internal/provider/providers/vultr/provider.go
```diff
@@ -0,0 +1,195 @@
...
+	case ttl == 0:
+		return fmt.Errorf("%w", errors.ErrTTLNotSet)
```
qdm12, line 69: ttl is not required in API requests, so allow it to be zero, and just don't send it with the request if it's zero.

internal/provider/providers/vultr/provider.go
```diff
@@ -0,0 +1,195 @@
...
+		Scheme: "https",
+		Host:   "api.vultr.com",
+		Path:   fmt.Sprintf("/v2/domains/%s/records", p.domain),
+	}
+
+	requestData := struct {
+		Type string `json:"type"`
+		IP   string `json:"data"`
+		Name string `json:"name"`
+		TTL  uint32 `json:"ttl"`
```
qdm12, line 134: So that the ttl field is not set if its value is zero.

```suggestion
		TTL  uint32 `json:"ttl,omitempty"`
```

internal/provider/providers/vultr/provider.go
```diff
@@ -0,0 +1,195 @@
...
+	err = decoder.Decode(&parsedJSON)
+	if err != nil {
+		return netip.Addr{}, fmt.Errorf("json decoding response body: %w", err)
+	}
+
+	if parsedJSON.Error != "" {
+		return netip.Addr{}, fmt.Errorf("API Error: %s", parsedJSON.Error)
+	}
+
+	if response.StatusCode != http.StatusCreated {
```
qdm12, line 182: Oh it's always statusCreated, never StatusOk (200)?

internal/provider/providers/vultr/provider.go
```diff
@@ -0,0 +1,195 @@
...
+		return netip.Addr{}, fmt.Errorf("%w: %d: %s",
+			errors.ErrHTTPStatusNotValid, response.StatusCode, utils.BodyToSingleLine(response.Body))
+	}
```
qdm12, line 185: I feel like this code would never trigger because we check the parsed error before it. Maybe have something like this instead:

```go
	if response.StatusCode != http.StatusCreated {
			errors.ErrHTTPStatusNotValid, response.StatusCode, parsedJSON.Error)
	}

	if parsedJSON.Error != "" {
		return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrUnsuccessful, parsedJSON.Error)
	}
```

Also for your own knowledge, note the previously used `utils.BodyToSingleLine(response.Body)` would not work because it has already been read by the JSON decoder. One way to work around this is to io.ReadAll(response.Body) to have it in a byte slice, then use `json.Unmarshal(byteSlice, &parsedJSON)` and you can then re-use `byteSlice` after for example with `utils.ToSingleLine(string(byteSlice))`. Not something to do here really I think though 😄

qdm12 (changes requested):
Hi there! Thanks for the PR and congrats on your first few Go steps! 💯 
Looking pretty good, with a few minor changes to make 😉

docs/vultr.md
amroessam: Fixing

docs/vultr.md
amroessam: Yes, I've tested it

docs/vultr.md
amroessam: Good shout, let me do that

internal/provider/providers/vultr/provider.go
amroessam: Yes sir ✅

internal/provider/providers/vultr/provider.go
amroessam: This is a bit confusing, the token is actually a bearer token, but from their settings page they're asking for an API key, I think APIKeyNotSet makes more sense as that's what the user is asked to send.

internal/provider/providers/vultr/provider.go
amroessam: I can't flip it as that would return, their api returns an error string if it's not 201, so that checks for the error message and displays it for the user. If the object doesn't contain an error though, we can rely on the status code. 

Thanks for your tip

.gitignore
amroessam: #835 handled here

internal/provider/providers/vultr/provider.go
amroessam: I've updated to be token not set, let me know what you prefer though 🙏🏽

amroessam: image

qdm12: Oh ok, let's match what vultr has, so instead rename that token field to `apikey` and use that previous apikey not set error. The user shouldn't (and probably doesn't) care about the implementation details such as how it's sent to the API (bearer token).

internal/provider/providers/vultr/provider.go
amroessam: There are 3 cases
1. 201 created
2. HTTP error (400, 401, 403, 404) with error message
3. No documentation about 5xx where probably there won't be an error message

So I first check the error message if it exists, and if not, I check the status code. Does that make sense?

internal/provider/providers/vultr/provider.go
amroessam: Fixed, take a look pls 🙏🏽

internal/provider/providers/vultr/provider.go
qdm12: Ah interesting, so the code needs more change then.

I suggest you have a little function:

```go
func parseJSONError(body io.Reader) (message string) {
	var parsedJSON struct {
		Error string `json:"error"`
	}
	decoder := json.NewDecoder(body)
	err := decoder.Decode(&parsedJSON)
	if err != nil {
		return "<json decoding error: " + err.Error() + ">"
	}
	return parsedJSON.Error
}
```

and then have a switch on the status code:


```go
	defer response.Body.Close()
	
	switch response.StatusCode {
	case http.StatusCreated:
	case http.StatusBadRequest:
		return fmt.Errorf("%w: %s", errors.ErrBadRequest, parseJSONError(response.Body))
	case http.StatusUnauthorized, http.StatusForbidden:
		return fmt.Errorf("%w: %s", errors.ErrAuth, parseJSONError(response.Body))
	case http.StatusNotFound:
		return fmt.Errorf("%w: %s", errors.ErrDomainNotFound, parseJSONError(response.Body))
	default:
		return netip.Addr{}, fmt.Errorf("%w: %s: %s",	errors.ErrHTTPStatusNotValid,
			response.Status, utils.BodyToSingleLine(response.Body))
	}

	decoder := json.NewDecoder(response.Body)
	var parsedJSON struct {
		Error string
		Status uint32
		Record struct {
			ID   string `json:"id"`
			Type string `json:"type"`
			IP   string `json:"data"`
			Name string `json:"name"`
			TTL  uint32 `json:"ttl"`
		}
	}
	err = decoder.Decode(&parsedJSON)
	if err != nil {
		return netip.Addr{}, fmt.Errorf("json decoding response body: %w", err)
	}
```

internal/provider/providers/vultr/provider.go
```diff
@@ -0,0 +1,190 @@
...
+	response, err := client.Do(request)
+	if err != nil {
+		return netip.Addr{}, err
+	}
+	defer response.Body.Close()
+
+	decoder := json.NewDecoder(response.Body)
+	var parsedJSON struct {
+		Error string
+		Status uint32
```
qdm12, line 159: Also what is that Status used for? 🤔

internal/provider/providers/vultr/provider.go
```diff
@@ -0,0 +1,190 @@
...
+	ipv6Suffix netip.Prefix
+	apiKey      string
+	ttl        uint32
+}
+
+func New(data json.RawMessage, domain, owner string,
+	ipVersion ipversion.IPVersion, ipv6Suffix netip.Prefix) (
+	provider *Provider, err error) {
+	var providerSpecificSettings struct {
+		APIKey string `json:"apiKey"`
```
qdm12, line 33:
```suggestion
		APIKey string `json:"apikey"`
```

docs/vultr.md
```diff
@@ -0,0 +1,30 @@
...
+
+### Example
+
+```json
+{
+  "settings": [
+    {
+      "provider": "vultr",
+      "domain": "potato.example.com",
+      "apiKey": "AAAAAAAAAAAAAAA",
```
qdm12, line 13: nit to match all-lowercase json elsewhere in this codebase

```suggestion
      "apikey": "AAAAAAAAAAAAAAA",
```

docs/vultr.md
```diff
@@ -0,0 +1,30 @@
...
+      "ip_version": "ipv4"
+    }
+  ]
+}
+```
+
+### Compulsory parameters
+
+- `"domain"` is the domain to update. It can be `example.com` (root domain) or `potato.example.com` (subdomain of `example.com`). In case of a wildcard, this only works if there are no existing wildcard records of any type. If there is, it should error out.
+- `"apiKey"` is your API key, can be obtained from the [account settings](https://my.vultr.com/settings/#settingsapi), this is used as a bearer token to authenticate your request
```
qdm12, line 24:
```suggestion
- `"apikey"` is your API key, can be obtained from the [account settings](https://my.vultr.com/settings/#settingsapi), this is used as a bearer token to authenticate your request
```

internal/provider/providers/vultr/provider.go
qdm12: Thanks! I also love how they say "Personal Access Token" and also "API key"... I guess it doesn't matter too much 😄

internal/provider/providers/vultr/provider.go
amroessam: It's the HTTP status code

internal/provider/providers/vultr/provider.go
amroessam: Sure, but why not camelCase? 🤔

internal/provider/providers/vultr/provider.go
qdm12: Ok that's weird 😄 Should we... remove it from that struct?? Since it's... well, in the HTTP response already...

amroessam:
Sorry for the delay, I was away. Can you recheck?

internal/provider/providers/vultr/provider.go
amroessam: For the sake of implementing the API, I've added it, can remove it if you insist 🙏🏽

internal/provider/providers/vultr/provider.go
qdm12: Just to match other existing fields, which are all snake case. Also a bit of a habit to have snake case json keys.

qdm12:
Hey there, thanks for the last few commits!
In the end I decided to push directly on your branch, since there were quite a few things to change (ranging from not important at all to somewhat important - see commits with `fix: ` prefix). Feel free to browse commit by commit, since I tried my best to separate changes one by one for knowledge sharing purposes!
I also rebased your branch on the master branch here, so feel free to get changes using `git checkout master; git fetch; git reset --hard origin/master` 😉 
Just waiting for the CI to pass and I'll merge this. Thanks!

PR #835, opened 2024-10-08
Update gitignore to ignore build artificats

amroessam:


.gitignore
```diff
@@ -1 +1,4 @@
-/data
\ No newline at end of file
+/data
+.vscode
+app
```
qdm12, line 3: where is `app` coming from though?

.gitignore
```diff
@@ -1 +1,4 @@
...
+cmd/ddns-updater/data/
```
qdm12, line 4: Why would the data directory be in `cmd/ddns-updater/data/`?
On my side, I usually build and run from the root of the workspace with `go run ./cmd/ddns-updater/main.go` and it picks up the `data` directory there.

amroessam: It should be ignored because while developing you would probably add your ddns secrets in there which would then be leaked if pushed

.gitignore
amroessam: That's the built binary, should be ignored in. my opinion

qdm12:
Closing as not really needed.

Issue #809, opened 2024-09-09
Feature request: Allow updating to Private IP address instead of Public IP address


MRDGH2821:
1. What's the feature?
As the title says, please add an option to allow for DDNS updates with private IP instead of public IP.

2. Extra information?
My network is under double NAT system, and thus nothing is accessible. So in DuckDNS I have put my device's private network local IP address (`192.168.1.80`).

I have made a DuckDNS updater bash script which can check if the public IP is pingable or not. If yes update to that, else use private IP. So I was thinking that if such functionality can come in DDNS-Updater too, that would be nice.

Here is my bash script for reference:
<detail>
```sh
#!/usr/bin/env bash

set -e

# Logging configuration
LOG_FILE="./duckdns_update.log"

# Function to log messages
log() {
    local message="$1"
    local log_message="$(date '+%Y-%m-%d %H:%M:%S') - $message"
    echo $message
    echo $log_message >>"$LOG_FILE"
}

# Function to get the private IP address
get_private_ip() {
    hostname -i | awk '{print $3}'
}

# Function to get the public IP address
get_public_ip() {
    curl -s https://api.ipify.org
}

# Function to check if an IP address is reachable
is_ip_reachable() {
    ping -c 1 "$1" &>/dev/null
    return $?
}

# Function to update DuckDNS
update_duckdns() {
    local ip="$1"
    local url="https://www.duckdns.org/update?domains=${DUCKDNS_DOMAIN}&token=${DUCKDNS_TOKEN}&ip=${ip}"
    response=$(curl -s "$url")
    log "DuckDNS update response: $response"
}

# Main script logic
public_ip=$(get_public_ip)

if [ -n "$public_ip" ] && is_ip_reachable "$public_ip"; then
    log "Public IP $public_ip is reachable. Updating DuckDNS..."
    update_duckdns "$public_ip"
else
    private_ip=$(get_private_ip)
    log "Public IP ($public_ip) not reachable. Updating DuckDNS with private IP $private_ip..."
    update_duckdns "$private_ip"
fi
```
</detail>

primez:
Hello @MRDGH2821,

I am not a maintainer of this project, but I also needed this feature and decided to implement it myself.
I am not sure if it will be accepted but you can download the code and use my version - it has been verified in a container and a Linux env.
Here is a PR https://github.com/qdm12/ddns-updater/pull/839

MRDGH2821:
> Hello @MRDGH2821,
> 
[...]

PR #839, opened 2024-10-12
[Feature] Added a possibility to use a private IP instead of a public IP

primez:
Hello,

First of all, thank you for such a great tool!

I decided to implement support for using private IPs because I needed this feature and noticed that the related feature request at https://github.com/qdm12/ddns-updater/issues/809 has been open for a month with no replies. I assumed it might not be a priority or not planned at all.

Disclosure: I am a .NET developer, and my knowledge of Golang is limited. I relied heavily on ChatGPT to help me write this code. However, I have verified its functionality in a container and a Linux environment, and it works as expected, binding a private IP to a domain name.

To make this work, I added a new value to the environment variables:
```
 environment:
      - PUBLICIP_FETCHERS=privateip
```


MRDGH2821:
If run in bare metal, is the private IP in 192.168.X.Y ?

If run inside container, does it give the same private IP as bare metal?

(Bare metal means directly installing on host system)

BTW, thanks for this PR!
I hope it will get accepted!

primez:
A bare metal installation will give you an IP in the range of 192.168..., as well as a VM installation.
In a container, well, I think it depends on a network. I haven't tested this extensively, but in the simplest case, it will give the IP of a container, not the host system.
So, I installed it as a Linux service that starts with the host OS.

Neustradamus:
@qdm12: What do you think?

Linked to:
- https://github.com/qdm12/ddns-updater/issues/809

Lawstorant:
I think it would be great if we could obtain the private IP from DNS as well. I have a weird setup where I would want to update my DDNS entry with the ip of the `serwer.lan` entry. `serwer` is my local server which uses DHCP to obtain it's IP.

Of course, I gave it a static, infinite lease BUT in case I ever change it, I'd like it to change automatically everywhere.

PR #846, opened 2024-10-21
fix(gcp): prevent crash for missing record

mcfedr:
Currently if the record is missing from GCP updater will crash with a null pointer at `for _, rrdata := range recordResourceSet.Rrdatas {`

so this moves the case when its null first and prevents the crash


qdm12 (approved):
Thanks for the fix!!! 🎖️

PR #869, opened 2024-11-25
Docs for using `ipv6_suffix` to update permanent IPV6 address

### LINKED ISSUE #868: Feature request: update `permanent` IPV6 address instead of `temporary` one
Currently, I find it register server's `temporary` IPV6 address instead of `permanent` address, making it dynamic regularly.

I wonder if I could use `permanent` IPV6 address instead. I'm using `cloudflare` provider without proxy.

ain-soph:
Close #868 

Docs for using `ipv6_suffix` to update permanent IPV6 address instead of temporary IPV6 address.

### DIFF
--- a/docs/cloudflare.md
+++ b/docs/cloudflare.md
@@ -37,4 +37,6 @@ See [this issue comment for context](https://github.com/qdm12/ddns-updater/issue
 - `"ip_version"` can be `ipv4` (A records), or `ipv6` (AAAA records) or `ipv4 or ipv6` (update one of the two, depending on the public ip found). It defaults to `ipv4 or ipv6`.
 - `"ipv6_suffix"` is the IPv6 interface identifier suffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw public IPv6 address obtained is used in the record updating.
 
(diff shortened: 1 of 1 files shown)


docs/cloudflare.md
```diff
@@ -37,4 +37,6 @@ See [this issue comment for context](https://github.com/qdm12/ddns-updater/issue
 - `"ip_version"` can be `ipv4` (A records), or `ipv6` (AAAA records) or `ipv4 or ipv6` (update one of the two, depending on the public ip found). It defaults to `ipv4 or ipv6`.
 - `"ipv6_suffix"` is the IPv6 interface identifier suffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw public IPv6 address obtained is used in the record updating.
 
+> Note: You could set `"ipv6_suffix"` with your permanent IPV6 address suffix (e.g., `0:0:0:0:72ad:8fbb:a54e:bedd/64` in previous example) to avoid updating the temporary IPV6 address.
```
qdm12, line 40: 🤔 I think it might be better to rephrase the bullet point above? Maybe:

- `"ipv6_suffix"` is the IPv6 interface identifier suffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw temporary IPv6 address of the machine is used in the record updating.

What do you think?? If so, we should extend this rephrasing to all providers document (it's the same bullet point word for word for all of them)

qdm12:
Done in 39bfd2e0d89a1ba6ac4c5de5facdc2eae4166b0d thanks!

PR #874, opened 2024-11-29
Use `RealIP` middleware

erri120:
Fixes logging issues when running behind a reverse proxy.

From the [docs](https://github.com/go-chi/chi?tab=readme-ov-file#core-middlewares):

> RealIP: Sets a http.Request's RemoteAddr to either X-Real-IP or X-Forwarded-For

Without this middleware, running the server behind a reverse proxy will result in every log entry to show the IP address of the proxy instead of the real IP address of the connecting client.


qdm12 (approved):
Thanks!

PR #885, opened 2024-12-22
Add myaddr dynamic dns provider

### LINKED ISSUE #728: `myaddr.tools`
1. What's the feature?

[myaddr.tools](https://myaddr.tools/).

2. Extra information?

Basically `GET`/`POST`/`PUT` `https://myaddr.tools/update?key=$key&ip=self` (subdomain ipv4/ipv6 exists for … forcing one of methods over there?).

While at it, there's yet another ip-address-detection service: [myip.addr.tools](https://myip.addr.tools/help).

brianshea2:
Adds new dynamic dns provider: https://myaddr.tools
Partially fixes #728 (additional PR for public IP provider to come)
Cheers!


cmd/ddns-updater/main.go
```diff
@@ -167,6 +167,14 @@ func _main(ctx context.Context, reader *reader.Reader, args []string, logger log
...
 	defer client.CloseIdleConnections()
 
+	for _, p := range providers {
+		err = provider.Init(p, ctx, client)
+		if err != nil {
+			shoutrrrClient.Notify(err.Error())
+			return err
+		}
+	}
+
```
qdm12, line 177: Let's not do this, just force the user to set a domain name in the config.json, to match the existing architecture. There are also issues where for example the addrtools API does not work, then we won't know what domain name is failing (just the key) which isn't great.

internal/params/json.go
```diff
@@ -171,11 +171,13 @@ func makeSettingsFromObject(common commonSettings, rawSettings json.RawMessage,
...
 			fmt.Sprintf("you can specify the owner %q directly in the domain field as %q",
 				common.Owner, strings.Join(domains, ",")))
-	} else { // extract owner(s) from domain(s)
+	} else if common.Domain != "" { // extract owner(s) from domain(s)
 		domain, owners, err = extractFromDomainField(common.Domain)
 		if err != nil {
 			return nil, nil, fmt.Errorf("extracting owners from domains: %w", err)
 		}
+	} else { // domain(s) not specified
+		owners = make([]string, 1) // required to enter provider initialization loop below
```
qdm12, line 180: Related to init: revert this once the init function is removed 😉

internal/provider/provider.go
```diff
@@ -75,6 +76,10 @@ type Provider interface {
 	Update(ctx context.Context, client *http.Client, ip netip.Addr) (newIP netip.Addr, err error)
 }
 
+type Initializer interface {
+	Init(ctx context.Context, client *http.Client) error
+}
+
```
qdm12, line 82: Related to init: remove this

internal/provider/provider.go
```diff
@@ -186,3 +193,10 @@ func New(providerName models.Provider, data json.RawMessage, domain, owner strin
 		return nil, fmt.Errorf("%w: %s", ErrProviderUnknown, providerName)
 	}
 }
+
+func Init(provider Provider, ctx context.Context, client *http.Client) error {
+	if initializer, ok := provider.(Initializer); ok {
+		return initializer.Init(ctx, client)
+	}
+	return nil
+}
```
qdm12, line 202: Related to init: remove

internal/provider/providers/myaddr/provider.go
```diff
@@ -0,0 +1,178 @@
...
+	ipv6Suffix netip.Prefix
+}
+
+func New(data json.RawMessage, ipVersion ipversion.IPVersion, ipv6Suffix netip.Prefix) (p *Provider, err error) {
+	var providerSpecificSettings struct {
+		Key string `json:"key"`
+	}
+	if err = json.Unmarshal(data, &providerSpecificSettings); err != nil {
+		err = fmt.Errorf("json decoding provider specific settings (myaddr): %w", err)
+		return
```
qdm12, line 34:
```suggestion
		return nil, fmt.Errorf("json decoding provider specific settings: %w", err)
```

Also no point having the provider name, that's wrapped by calling layers (or should be, if not!)

internal/provider/providers/myaddr/provider.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	if providerSpecificSettings.Key == "" {
+		err = fmt.Errorf("validating provider specific settings (myaddr): %w", errors.ErrKeyNotSet)
+		return
```
qdm12, line 38:
```suggestion
		return nil, fmt.Errorf("validating provider specific settings: %w", errors.ErrKeyNotSet)
```

internal/provider/providers/myaddr/provider.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	p = &Provider{
+		key:        providerSpecificSettings.Key,
+		ipVersion:  ipVersion,
+		ipv6Suffix: ipv6Suffix,
+	}
+	return
```
qdm12, line 45:
```suggestion
	return &Provider{
		key:        providerSpecificSettings.Key,
		ipVersion:  ipVersion,
		ipv6Suffix: ipv6Suffix,
	}, nil
```

internal/provider/providers/myaddr/provider.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func (p *Provider) Init(ctx context.Context, client *http.Client) (err error) {
```
qdm12, line 48: remove entire function 🙏 Thanks for trying to optimize this though!

internal/provider/providers/myaddr/provider.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	v := url.Values{}
+	v.Set("key", p.key)
+	v.Set("ip", ip.String())
+	b := strings.NewReader(v.Encode())
+	var req *http.Request
+	if req, err = http.NewRequestWithContext(ctx, http.MethodPost, u.String(), b); err != nil {
+		err = fmt.Errorf("creating http request: %w", err)
+		return
+	}
```
qdm12, line 149: no short ifs (more readable), no naked returns (otherwise it's hard to know what's being returned), and let's use full name variable names `req` -> `request` for better readability

```suggestion
	request, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), b)
	if err != nil {
		return netip.Addr{}, fmt.Errorf("creating http request: %w", err)
	}
```

internal/provider/providers/myaddr/provider.go
```diff
@@ -0,0 +1,178 @@
...
+	headers.SetContentType(req, "application/x-www-form-urlencoded")
+	headers.SetUserAgent(req)
+
+	// execute HTTP request
```
qdm12, line 153: unneeded comment, the code speaks for itself below

```suggestion
```

internal/provider/providers/myaddr/provider.go
```diff
@@ -0,0 +1,178 @@
...
+	var resp *http.Response
+	if resp, err = client.Do(req); err != nil {
+		return
+	}
```
qdm12, line 157:
```suggestion
	response, err := client.Do(request)
	if err != nil {
		return netip.Addr{}, err
	}
```

internal/provider/providers/myaddr/provider.go
```diff
@@ -0,0 +1,178 @@
...
+	defer resp.Body.Close()
+	defer io.Copy(io.Discard, resp.Body) // ensure body is read to completion
```
qdm12, line 159: Interesting, but unneeded, see https://gist.github.com/mholt/eba0f2cc96658be0f717?permalink_comment_id=2605879#gistcomment-2605879

```suggestion
```

internal/provider/providers/myaddr/provider.go
```diff
@@ -0,0 +1,178 @@
...
+		err = fmt.Errorf("%w: %s", errors.ErrBadRequest, utils.BodyToSingleLine(resp.Body))
+	case http.StatusNotFound:
+		// registration not found
+		err = fmt.Errorf("%w: %s", errors.ErrKeyNotValid, utils.BodyToSingleLine(resp.Body))
+	default:
+		// unexpected HTTP status
+		err = fmt.Errorf("%w: %d: %s",
+			errors.ErrHTTPStatusNotValid, resp.StatusCode, utils.BodyToSingleLine(resp.Body))
+	}
+	return
```
qdm12, line 177: - some unneeded comments
- no naked returns
- return early, almost always, easier to read

```suggestion
	switch response.StatusCode {
	case http.StatusOK:
		return ip, nil
	case http.StatusBadRequest:
		return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrBadRequest, utils.BodyToSingleLine(response.Body))
	case http.StatusNotFound
		return netip.Addr{}, fmt.Errorf("%w: %s", errors.ErrKeyNotValid, utils.BodyToSingleLine(response.Body))
	default:
		return netip.Addr{}, fmt.Errorf("%w: %d: %s",
			errors.ErrHTTPStatusNotValid, response.StatusCode, utils.BodyToSingleLine(response.Body))
	}
```

internal/provider/providers/myaddr/provider.go
```diff
@@ -0,0 +1,178 @@
...
+	// generate HTTP request
+	u := &url.URL{
+		Scheme: "https",
+		Host:   "myaddr.tools",
+		Path:   "/update",
+	}
+	v := url.Values{}
+	v.Set("key", p.key)
+	v.Set("ip", ip.String())
+	b := strings.NewReader(v.Encode())
```
qdm12, line 144: nit `b` -> `buffer` would be nice 😉

internal/provider/providers/myaddr/provider.go
```diff
@@ -0,0 +1,178 @@
...
+func (p *Provider) IPVersion() ipversion.IPVersion {
+	return p.ipVersion
+}
+
+func (p *Provider) IPv6Suffix() netip.Prefix {
+	return p.ipv6Suffix
+}
+
+func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Addr) (newIP netip.Addr, err error) {
+	// generate HTTP request
```
qdm12, line 135: uneeded comment

```suggestion
```

qdm12 (changes requested):
Thanks for the PR 🎖️ 💯 !

A few comments of things to change here and there, but before jumping in this: would you mind creating another separate pull request for the public ip feature, since it should be a separate commit/PR really. Thank you 🙏 👍

brianshea2:
Thank you for the quick review :) I believe I've addressed all your comments and made the code more consistent with the project.

I've reverted the public IP provider addition for now and will open a separate PR for that.

I understand your want for providing the `"domain"` to be consistent with other providers, and have removed the initialization code which fetched the domain corresponding to the provided `"key"`. My concern now is there's nothing explicitly linking the provided `"key"` and `"domain"`. I believe I've described the `"domain"` adequately in docs/myaddr.md to this point. But to reiterate, the provided `"key"` is the only identifier sent to the provider api, and is used to update all `[*.]name-corresponding-to-key.myaddr.{tools,dev,io}` in one fell swoop, which may or may not agree with the separately-user-provided `"domain"`, now used only for checking if an update is required. What are your thoughts here? It's possible to perform an additional api request during each update to fetch the domain corresponding to `"key"` and compare to the provided `"domain"` if wanted.

qdm12 (approved):

qdm12:
> I believe I've addressed all your comments and made the code more consistent with the project.

Awesome 💯 

> I've reverted the public IP provider addition for now and will open a separate PR for that.

Perfect 🥇 

> It's possible to perform an additional api request during each update to fetch the domain corresponding to "key" and compare to the provided "domain" if wanted.

Eh it's a bit on the user if he/she sets a key not matching the domain. We can do foolproof checks as much as possible, but this seems like a bit too much at the cost of over-complicating it 😄 The API will probably just say the key is wrong anyway.

I did however add a little check to prevent multiple domains to be defined in a myaddr settings block since one key is for one domain (and its subdomains), so there would be no point handling it as different entries/updates in parallel, except getting banned for sending too many update requests 😄 

Merging this !!

PR #889, opened 2024-12-24
Add addrtools public IP provider

brianshea2:
Adds new public IP provider: https://myip.addr.tools/help
Completes #728 (along with #885)
Cheers!


brianshea2:
The Markdown check is comically failing at checking the IPv6-only link for this new public IP provider. Rest assured, the link is correct :)

PR #914, opened 2025-01-18
Request to update cloudflare.md documentation

salocinyeet:
Added the "host" parameters to inform users how to update subdomains.
Putting a subdomain in the domain parameters, ended adding an additional period when configured.
For example, when domain was set to "sub.example.com", DDNS-Updater will attempt to update ".sub.example.com" which ended up with an error.

Feel free to ask questions

### DIFF
--- a/docs/cloudflare.md
+++ b/docs/cloudflare.md
@@ -23,7 +23,7 @@
 ### Compulsory parameters
 
 - `"zone_identifier"` is the Zone ID of your site, from the domain overview page written as *Zone ID*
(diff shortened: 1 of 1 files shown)


qdm12:
Users should NOT use the host parameter anymore, it's just kept for retrocompatibility. The host (aka owner) is automatically extracted from the domain field. The prefix dot you've seen is either a misconfiguration on your side or a code bug which needs to be fixed. Please try again, and, if it prefixes a dot in front then either open an issue or open a PR to fix the code (but really code should work fine, that part is tested quite a bit).

PR #921, opened 2025-01-25
feat(provider): add transip.nl

### LINKED ISSUE #254: `transip.nl`
www.transip.nl

LucaScorpion:
Adds support for [TransIP](https://www.transip.nl/), resolves #254

Tested locally, and seemed to work like a charm! Ipv6 and wildcards are both supported, and it will either create a new entry, or update an existing one.

The relevant API documentation can be found here: https://api.transip.nl/rest/docs.html#domains-dns

Side note: this PR also includes a small fix from https://github.com/qdm12/ddns-updater/pull/918. It shouldn't conflict, but if that does happen I'll fix that after that one is merged :)

Let me know if any changes are needed, and thanks in advance!


internal/provider/providers/transip/provider.go
```diff
@@ -0,0 +1,285 @@
...
+	// Remove the begin and end markers, remove whitespace, and trim.
+	pemData := strings.ReplaceAll(keyString, "\n", "")
+	pemData = strings.ReplaceAll(pemData, "-----BEGIN PRIVATE KEY-----", "")
+	pemData = strings.ReplaceAll(pemData, "-----END PRIVATE KEY-----", "")
+	pemData = strings.TrimSpace(pemData)
+
+	decodedKey, err := base64.StdEncoding.DecodeString(pemData)
+	if err != nil {
+		return nil, err
+	}
```
GeertJohan, line 131: Would it perhaps be better to let the `encoding/pem` package do this? That makes the parsing more compliant with the PEM standard.

```suggestion
	block, _ := pem.Decode(pubPEMData)
	if block == nil || block.Type != "PRIVATE KEY" {
		return nil, errors.New("failed to decode PEM block containing private key")
	}
```

And below substitute `decodedKey` for `block.Bytes`.

GeertJohan:
I've tested this PR and it seems to work good :) 
If anyone wants to use it; a container image with this feature built in is available at `docker.io/geertjohan/ddns-updater:with-transip`.

LucaScorpion:
@GeertJohan Thanks for testing and the suggestion! I'll try to update the PR somewhere next week to include the suggestion and get it fully ready :)

LucaScorpion:
Took a bit longer than expected, but the suggestion is now part of the code, and all seems to be working well still :)

GeertJohan:
I have just pushed a new build to `docker.io/geertjohan/ddns-updater:with-transip` and am using it succesfully.

docs/transip.md
```diff
@@ -0,0 +1,42 @@
...
+```json
+{
+  "settings": [
+    {
+      "provider": "transip",
+      "domain": "example.com",
+      "ip_version":  "ipv4",
+      "ipv6_suffix": "",
+      "username": "username",
+      "key": "-----BEGIN PRIVATE KEY-----...-----END PRIVATE KEY-----"
```
qdm12, line 16: Is this given as a single line from transip?

docs/transip.md
```diff
@@ -0,0 +1,42 @@
...
+### Compulsory parameters
+
+- `"domain"` is the domain to update. It can be `example.com` (root domain), `sub.example.com` (subdomain of `example.com`) or `*.example.com` for the wildcard.
+- `"username"`
+- `"key"`
+
+### Optional parameters
+
+- `"ip_version"` can be `ipv4` (A records), or `ipv6` (AAAA records) or `ipv4 or ipv6` (update one of the two, depending on the public ip found). It defaults to `ipv4 or ipv6`.
+- `"ipv6_suffix"` is the IPv6 interface identifier suffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw public IPv6 address obtained is used in the record updating.
```
qdm12, line 31: (from master branch)

```suggestion
- `"ipv6_suffix"` is the IPv6 interface identifier suffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw temporary IPv6 address of the machine is used in the record updating. You might want to set this to use your permanent IPv6 address instead of your temporary IPv6 address.
```

docs/transip.md
```diff
@@ -0,0 +1,42 @@
...
+## Domain setup
+
+1. Log in to the [TransIP control panel](https://www.transip.nl/cp/).
+2. Under your account, go to the [API settings](https://www.transip.nl/cp/account/api/).
+3. Enable the API.\
+![A toggle showing the API status as on](../readme/transip1.png)
+4. Add a key pair. Make sure to uncheck the checkbox to only accept IP addresses from the whitelist.\
+![A table listing the key pairs](../readme/transip2.png)
+5. Copy your private key, and store it in your config file.\
+![A snippet of a generated private key](../readme/transip3.png)
```
qdm12, line 42:
```suggestion
3. Enable the API.

    ![A toggle showing the API status as on](../readme/transip1.png)

4. Add a key pair. Make sure to uncheck the checkbox to only accept IP addresses from the whitelist.

    ![A table listing the key pairs](../readme/transip2.png)

5. Copy your private key, and store it in your config file.

    ![A snippet of a generated private key](../readme/transip3.png)
```

internal/provider/providers/transip/provider.go
```diff
@@ -0,0 +1,279 @@
...
+		return fmt.Errorf("%w: %w", errors.ErrDomainNotValid, err)
+	}
+
+	if username == "" {
+		return errors.ErrUsernameNotSet
+	}
+
+	if key == "" {
+		return errors.ErrKeyNotSet
+	}
```
qdm12, line 75: use switch for less LoC and wrap errors to enforce calling layers use `errors.Is` and not `==` to check errors.

```suggestion
	switch {
	case username == "":
		return fmt.Errorf("%w", errors.ErrUsernameNotSet)
	case key == "":
		return fmt.Errorf("%w", errors.ErrKeyNotSet)
	}
```

internal/provider/providers/transip/provider.go
```diff
@@ -0,0 +1,279 @@
...
+		Owner:     p.Owner(),
+		Provider:  "<a href=\"https://www.transip.nl/\">TransIP</a>",
+		IPVersion: p.ipVersion.String(),
+	}
+}
+
+func parsePrivateKey(keyString string) (*rsa.PrivateKey, error) {
+	block, _ := pem.Decode([]byte(keyString))
+	if block == nil || block.Type != "PRIVATE KEY" {
+		return nil, fmt.Errorf("failed to decode PEM block containing private key")
```
qdm12, line 124: define a sentinel error (aka global error variable `var errXYZ = errors.New("xyz")`), or the linter will fail

internal/provider/providers/transip/provider.go
```diff
@@ -0,0 +1,279 @@
...
+	key, err := x509.ParsePKCS8PrivateKey(block.Bytes)
+	if err != nil {
+		return nil, err
+	}
+
+	if rsaKey, ok := key.(*rsa.PrivateKey); ok {
+		return rsaKey, nil
+	}
+
+	return nil, fmt.Errorf("not an RSA private key")
```
qdm12, line 136: need to have a sentinel error i.e. `errKeyNotRSA = errors.New("key is not an RSA key")` and then wrap it with `fmt.Errorf("%w", errKeyNotRSA)`

internal/provider/providers/transip/provider.go
```diff
@@ -0,0 +1,279 @@
...
+		"nonce":      strconv.FormatInt(time.Now().UnixNano(), 10),
+		"global_key": true,
+		"read_only":  false,
+		"label":      fmt.Sprintf("ddns-updater %d", time.Now().Unix()),
+	})
+	if err != nil {
+		return "", fmt.Errorf("json encoding request body: %w", err)
+	}
+
+	request, err := http.NewRequestWithContext(ctx, http.MethodPost, "https://api.transip.nl/v6/auth", bytes.NewReader(requestBody))
```
qdm12, line 151:
```suggestion
	buffer := bytes.NewBuffer(nil)
	encoder := json.NewEncoder(buffer)
	requestData := struct {
		Login		string	`json:"login"`
		Nonce		string	`json:"nonce"`
		GlobalKey	bool	`json:"global_key`
		ReadOnly	bool	`json:"readonly`
		Label		string	`json:"label"`
	}{
		Login: p.username,
		Nonce: strconv.FormatInt(time.Now().UnixNano(), 10),
		GlobalKey: true,
		ReadOnly: false,
		Label: fmt.Sprintf("ddns-updater %d", time.Now().Unix()),
	}
	err = encoder.Encode(requestData)
	if err != nil {
		return netip.Addr{}, fmt.Errorf("json encoding request data: %w", err)
	}

	const url = "https://api.transip.nl/v6/auth"
	request, err := http.NewRequestWithContext(ctx, http.MethodPost, url, buffer)
```

internal/provider/providers/transip/provider.go
```diff
@@ -0,0 +1,279 @@
...
+	if err != nil {
+		return "", fmt.Errorf("creating http request: %w", err)
+	}
+	headers.SetUserAgent(request)
+	headers.SetContentType(request, "application/json")
+
+	privateKey, err := parsePrivateKey(p.key)
+	if err != nil {
+		return "", fmt.Errorf("parsing private key: %w", err)
+	}
```
qdm12, line 161: let's keep a parsed private key in the provider struct, no need to parse it multiple times, just once when creating the provider

internal/provider/providers/transip/provider.go
```diff
@@ -0,0 +1,279 @@
...
+
+	// Sign the request body, put the signature in a header.
```
qdm12, line 163: nit unneeded comment


```suggestion
```

internal/provider/providers/transip/provider.go
```diff
@@ -0,0 +1,279 @@
...
+	hashedBody := sha512.Sum512(requestBody)
```
qdm12, line 164: nit rename `hashedBody` to `digest`

internal/provider/providers/transip/provider.go
```diff
@@ -0,0 +1,279 @@
...
+			errors.ErrHTTPStatusNotValid, response.StatusCode, utils.BodyToSingleLine(response.Body))
+	}
+
+	var result struct {
+		Token string `json:"token"`
+	}
+	err = json.NewDecoder(response.Body).Decode(&result)
+	if err != nil {
+		return "", fmt.Errorf("json decoding response body: %w", err)
+	}
```
qdm12, line 188: nit adding a check to make sure the token is found and not empty


```suggestion
	} else if result.Token == "" {
		return "", fmt.Errorf("in response body: %w", errors.ErrTokenNotSet)
	}
```

internal/provider/providers/transip/provider.go
```diff
@@ -0,0 +1,279 @@
...
+		return netip.Addr{}, fmt.Errorf("%w: %d: %s",
+			errors.ErrHTTPStatusNotValid, response.StatusCode, utils.BodyToSingleLine(response.Body))
+	}
+
+	return ip, nil
+}
+
+type dnsEntry struct {
+	Name    string `json:"name"`
+	Expire  int    `json:"expire"`
```
qdm12, line 276: omitempty so that if the user sets the ttl to zero, it's not used.

```suggestion
	Expire  uint    `json:"expire,omitempty"`
```

internal/provider/providers/transip/provider.go
```diff
@@ -0,0 +1,279 @@
...
+		return netip.Addr{}, fmt.Errorf("%w: %d: %s",
+			errors.ErrHTTPStatusNotValid, response.StatusCode, utils.BodyToSingleLine(response.Body))
+	}
+	var entries struct {
+		DnsEntries []dnsEntry `json:"dnsEntries"`
+	}
+	err = json.NewDecoder(response.Body).Decode(&entries)
+	if err != nil {
+		return netip.Addr{}, fmt.Errorf("json decoding response body: %w", err)
+	}
```
qdm12, line 228: move this to its own `getRecords` method

internal/provider/providers/transip/provider.go
```diff
@@ -0,0 +1,279 @@
...
+
+	// Construct the basis of the new/updated entry.
+	updatedEntry := dnsEntry{
+		Name:    p.owner,
+		Expire:  300,
```
qdm12, line 233: set this to a user defined ttl parameter, which, if `0`, is just not written so we keep the ttl already present (see below dnsEntry struct)

internal/provider/providers/transip/provider.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	// Create or update the entry.
+	requestBody, err := json.Marshal(map[string]any{
+		"dnsEntry": updatedEntry,
+	})
+	if err != nil {
+		return netip.Addr{}, fmt.Errorf("json encoding request body: %w", err)
+	}
+	request, err = http.NewRequestWithContext(ctx, postOrPatch, dnsApiUrl, bytes.NewReader(requestBody))
```
qdm12, line 254: I would prefer to have this split as a `createRecord` and an `updateRecord` method, eventhough they are very similar. That way you can go through records, if you find one, updateRecord, and, if not found, createRecord

internal/update/service.go
```diff
@@ -66,24 +65,19 @@ func (s *Service) lookupIPsResilient(ctx context.Context, hostname string, tries
...
 					continue // retry
 				}
 				results <- result{network: network, ips: ips, err: err}
 				return
 			}
-			results <- result{
-				network: network,
-				err:     fmt.Errorf("ip look up failed after %d tries: %w", tries, errors.Join(errs...)),
-			}
+			results <- result{network: network} // retries exceeded
```
qdm12, line 80: this can be removed, it's part of the master branch now 😉 Thanks still

qdm12 (changes requested):
Thanks!

For the png images, can you have them as compressed jpeg so they take less space (with a small loss of quality that is very much worth it in this case 😉), thanks!

docs/transip.md
```diff
@@ -0,0 +1,42 @@
...
+
+### Compulsory parameters
+
+- `"domain"` is the domain to update. It can be `example.com` (root domain), `sub.example.com` (subdomain of `example.com`) or `*.example.com` for the wildcard.
+- `"username"`
+- `"key"`
+
+### Optional parameters
+
+- `"ip_version"` can be `ipv4` (A records), or `ipv6` (AAAA records) or `ipv4 or ipv6` (update one of the two, depending on the public ip found). It defaults to `ipv4 or ipv6`.
```
qdm12, line 30: ttl should be added as an optional parameter.

PR #947, opened 2025-02-28
feat(provider): `nextdns`

### LINKED ISSUE #321: Feature request: add NextDNS.io provider
1. What's the feature?  Add the ability to update NextDNS.io via the DDNS updater.

2. Extra information?  NextDNS allows the ability to update IP address information via a single url by programmatically calling the following url: https://link-ip.nextdns.io/<Endpoint ID>/<API GUID>.  There is no need to provide a username/password or server.  Just need to call the URL to automatically update the IP.

hieunlt:
Adds support for [NextDNS](https://nextdns.io/), resolves https://github.com/qdm12/ddns-updater/issues/321


hieunlt:
@qdm12 can you take a look at my PR?

docs/nextdns.md
```diff
@@ -0,0 +1,40 @@
...
+      "ip_version": "ipv4",
+      "ipv6_suffix": ""
+    }
+  ]
+}
+```
+
+### Compulsory parameters
+
+- `"domain"` is the domain to update. For now, it must be "link-ip.nextdns.io".
```
qdm12, line 23: Does it always HAVE TO be `link-ip.nextdns.io`? If so, let's hardcode it and enforce `domain` is always set to this.

docs/nextdns.md
```diff
@@ -0,0 +1,40 @@
...
+- `"endpoint"`
```
qdm12, line 24: Should we split endpoint following the `https://link-ip.nextdns.io/<Endpoint ID>/<API GUID>` format from the issue? 🤔  It would be nice to add a bit of a description to this too.

docs/nextdns.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Optional parameters
+
+- `"ip_version"` can be `ipv4` (A records), or `ipv6` (AAAA records) or `ipv4 or ipv6` (update one of the two, depending on the public ip found). It defaults to `ipv4 or ipv6`.
+- `"ipv6_suffix"` is the IPv6 interface identifier suffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw public IPv6 address obtained is used in the record updating.
```
qdm12, line 29: (updating to master branch)

```suggestion
- `"ipv6_suffix"` is the IPv6 interface identifier suffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw temporary IPv6 address of the machine is used in the record updating. You might want to set this to use your permanent IPv6 address instead of your temporary IPv6 address.
```

docs/nextdns.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Domain setup
+
+> NextDNS supports updating Linked IP via a DDNS hostname. If you're already using a DDNS service, configure your DDNS domain in the Linked IP card instead.
+
+- Create an account on the [nextdns website](https://nextdns.io/)
+- Go to your [account page](https://my.nextdns.io/), login and setup Linked IP
+- Click `Show advanced options` button and copy the endpoint from the Linked IP card
+- Update the configuration file with the endpoint
```
qdm12, line 38:
```suggestion
1. Create an account on the [nextdns website](https://nextdns.io/)
1. Go to your [account page](https://my.nextdns.io/), login and setup Linked IP
1. Click `Show advanced options` button and copy the endpoint from the Linked IP card
1. Update the configuration file with the endpoint
```

docs/nextdns.md
```diff
@@ -0,0 +1,40 @@
...
+
+_See the [nextdns website](https://nextdns.io/)_
```
qdm12, line 40: nit unneeded formatting

```suggestion
See the [nextdns website](https://nextdns.io/)
```

docs/nextdns.md
```diff
@@ -0,0 +1,40 @@
...
+- `"endpoint"`
+
+### Optional parameters
+
+- `"ip_version"` can be `ipv4` (A records), or `ipv6` (AAAA records) or `ipv4 or ipv6` (update one of the two, depending on the public ip found). It defaults to `ipv4 or ipv6`.
+- `"ipv6_suffix"` is the IPv6 interface identifier suffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw public IPv6 address obtained is used in the record updating.
+
+## Domain setup
+
+> NextDNS supports updating Linked IP via a DDNS hostname. If you're already using a DDNS service, configure your DDNS domain in the Linked IP card instead.
```
qdm12, line 33: nit unneeded formatting

```suggestion
NextDNS supports updating Linked IP via a DDNS hostname. If you're already using a DDNS service, configure your DDNS domain in the Linked IP card instead.
```

internal/provider/providers/nextdns/provider.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	switch {
+	case !strings.HasSuffix(domain, "nextdns.io"):
+		return fmt.Errorf(`%w: %q must end with "%s"`,
+			errors.ErrDomainNotValid, domain, "nextdns.io")
+	case owner == "*":
+		return fmt.Errorf("%w: %s", errors.ErrOwnerWildcard, owner)
+	case !endpointRegex.MatchString(endpoint):
+		return fmt.Errorf("%w: endpoint %q does not match regex %q",
+			errors.ErrTokenNotValid, endpoint, endpointRegex)
```
qdm12, line 75: `errors.ErrTokenNotValid` is not appropriate here

internal/provider/providers/nextdns/provider.go
```diff
@@ -0,0 +1,144 @@
...
+		return netip.Addr{}, fmt.Errorf("creating http request: %w", err)
+	}
+
+	response, err := client.Do(request)
+	if err != nil {
+		return netip.Addr{}, err
+	}
+	defer response.Body.Close()
+
+	s, err := utils.ReadAndCleanBody(response.Body)
```
qdm12, line 134: What's in the response on success? How about with the wrong endpoint?

docs/nextdns.md
hieunlt: done

docs/nextdns.md
hieunlt: agree, also update the `Domain setup` section to guide users

internal/provider/providers/nextdns/provider.go
hieunlt: created `errors.ErrEndpointNotValid`

internal/provider/providers/nextdns/provider.go
hieunlt: Tested the `link-ip.nextdns.io` endpoint:

200 OK — success, plain-text OK body (IP updated)
400 Bad Request — plain-text Bad Request; happens when endpoint_id or api_guid don't match the expected format
404 Not Found — JSON {"errors":[{"code":"notFound"}]}; happens when path segments are missing entirely

hieunlt:
hi @qdm12 I've made a few adjustments based on your comments, can you take a look when you have time?

docs/nextdns.md
```diff
@@ -0,0 +1,44 @@
...
+      "ip_version": "ipv4",
+      "ipv6_suffix": ""
+    }
+  ]
+}
+```
+
+### Compulsory parameters
+
+- `"domain"` must be `"nextdns.io"` and `"owner"` must be `"link-ip"`.
```
qdm12: 1. let's stick to just using `"domain": "link-ip.nextdns.io"`. `owner` is just here for retrocompatibility
2. why have this field there if it's always a constant? Just have it as a constant in the code right? Maybe the code is too rigid to allow that, just saying this from the top of my head. If so, just change it to be `domain` only, without `owner`.

docs/nextdns.md
```diff
@@ -0,0 +1,44 @@
...
+
+NextDNS supports updating Linked IP via a DDNS hostname. If you're already using a DDNS service, configure your DDNS domain in the Linked IP card instead.
+
+1. Create an account on the [nextdns website](https://nextdns.io/)
+1. Go to your [account page](https://my.nextdns.io/), login and setup Linked IP
+1. Click `Show advanced options` in the Linked IP card
+1. Copy the update URL shown — it looks like `https://link-ip.nextdns.io/abcdef/0123456789abcdef`
+1. Set `"endpoint_id"` to the first path segment (e.g. `abcdef`) and `"api_guid"` to the second path segment (e.g. `0123456789abcdef`)
+
+See the [nextdns website](https://nextdns.io/)
```
qdm12: nit not necessary given instructions above 😉 

```suggestion

```

qdm12:
I did push a few changes, but importantly the `domain` field is now your actual domain name you are updating, not `link-ip.nextdns.io`. The update URL is constant in the update code. The domain field should be set by the user so that ddns-updater can dns resolve it and figure out when a dns update call is needed. This is critical to avoid updating every period, even if there is no need for it. Thoughts?

Edit: also apologies for the mega late replies, I'm swamped with life and code, still there, but lagging pretty hard

hieunlt:
> The domain field should be set by the user so that ddns-updater can dns resolve it and figure out when a dns update call is needed. This is critical to avoid updating every period, even if there is no need for it. Thoughts?

thanks man, as a follow up I removed redundant test cases, at first I believed that `domain` is fixed from NextDNS side, that's why I hardcoded it in the first place and thank you very much for spending time explain the mechanism to me 🙏

internal/provider/providers/nextdns/provider_test.go
```diff
@@ -11,58 +11,16 @@ import (
...
-				assert.EqualError(t, err, testCase.errMessage)
-				assert.Nil(t, provider)
-			} else {
-				require.NoError(t, err)
-				assert.NotNil(t, provider)
-			}
-		})
-	}
+	require.NoError(t, err)
+	assert.NotNil(t, provider)
```
qdm12: let's assert it against an expected Provider struct, otherwise this test is a bit useless except to check it does not error

qdm12 (approved):

PR #1025, opened 2025-08-09
Route53 Signer and Provider

enriquemanuel:
- Move the Signer to use AWS SDK
- Allow to specify AWS_PROFILE or Keys
- If using aws_profile it can be with SSO and Session Tokens
- Show the role and or profile being used in logs


internal/provider/providers/route53/provider.go
```diff
@@ -1,20 +1,19 @@
...
 	"context"
 	"encoding/json"
-	"encoding/xml"
 	"fmt"
 	"net/http"
 	"net/netip"
-	"net/url"
-	"time"
+
+	"github.com/aws/aws-sdk-go/aws/credentials"
```
simonmarty, line 10: Go SDK v1 is EOL https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/

You're gonna want to use `github.com/aws/aws-sdk-go-v2/`

qdm12:
That was previously dependency free, I expect for it to stay dependency free, and feel free to add features to the existing provider code, but without importing huge sdk dependencies.

Issue #1029, opened 2025-08-19
Bug: Internet dropout skips update and never retries


davidlang42:
**TLDR**: When the internet connection drops out, the update gets skipped and never tried again until the next change.

1. Is this urgent: YES
2. DNS provider(s) you use: DuckDNS
3. Program version:

`Running version 2.9.0 built on 2024-12-24T10:40:42.120Z (commit 78f3061)`

4. What are you using to run the container: TrueNAS "Apps" (aka TrueCharts)
5. Extra information: I've experienced this repeatedly, when the internet drops out, it fails to send an update, and then doesn't try again unless the container is stopped and restarted. What's worse, once I figured this out I saw in the logs that it only WARNS that it's doing this. It should probably be error? And also retry. The container crashing would even be better cause at least then I'd know there was a problem before I found myself unable to connect.

Logs:

```log
2025-08-15 08:20:15.375777+00:002025-08-15T18:20:15+10:00 ERROR obtaining ipv4 or ipv6 address: after 3 tries, errors were: Get "https://api.seeip.org": context deadline exceeded, dial tcp: lookup dns.opendns.com: i/o timeout, Get "https://ip.changeip.com": context deadline exceeded
2025-08-15 08:20:15.375902+00:002025-08-15T18:20:15+10:00 WARN Skipping update for REDACTED.duckdns.org because IP address was not found
```

Configuration file:

```json
{
  "settings": [
    {
      "domain": "duckdns.org",
      "ip_version": "",
      "provider": "duckdns",
      "token": "REDACTED"
    }
  ]
}
```

Host OS: TrueNAS Scale 24.20.2

tadeusz-niemiec:
Bump, this is crucial, I've just found myself unable to connect to my homelab. Since proper DNS records are required even for VPNs to work, whenever this bug occurs I'm stuck and can't even fix it remotely

tadeusz-niemiec:
EDIT: I'm sorry, looks like the thing I pasted was actually caused by my pihole instance blocking dns requests for some of the services used to obtain the IP address. Not sure why it works after a restart but definitely IPs resolved to 0.0.0.0 is pihole's fault

--- Original message ---

Hey, it happened to me again and I had to restart the container. As you can see immediately after restarting it everything worked fine (even though there's still some error visible at the very end).

What's suprising is that the last log of the previous run ends at `2026-01-27T11:26:06Z` with nothing else up until the restart next day.

```
========================================
========================================
[...]

PR #1042, opened 2025-10-17
Bump number of retries before abandoning all polling for a domain

directhex:
This should help weather the storm on brief connection outages. It's a bad and stupid solution, but it's less bad than the existing default.

Closes: https://github.com/qdm12/ddns-updater/issues/1029


qdm12:
Better fix brought through #1099 instead, thanks still!

PR #1052, opened 2025-11-05
feat: Added Netlify as a provider. 

CompewterTutor:
Added the provider (note…updates aren't supported so we have to delete and then recreate)., created documentation for it and updated the readme. Passes linting though I did have to migrate the lint config to get it to run but those changes are not included in this commit.

### DIFF
--- a/docs/netlify.md
+++ b/docs/netlify.md
@@ -1,0 +1,76 @@
+# Netlify
+
+## Configuration
+
+### Example
+
+```json
+{
+  "settings": [
+    {
+      "provider": "netlify",
+      "domain": "domain.com",
+      "token": "your-netlify-access-token",
+      "ip_version": "ipv4",
+      "ipv6_suffix": ""
+    }
+  ]
+}
+```
+
+### Compulsory parameters
+
+- `"domain"` is the domain to update. It can be `example.com` (root domain), `sub.example.com` (subdomain of `example.com`) or `*.example.com` for wildcard.
+- `"token"` is your Netlify personal access token with DNS zone permissions
+
+### Optional parameters
+
+- `"ip_version"` can be `ipv4` (A records), or `ipv6` (AAAA records) or `ipv4 or ipv6` (update one of the two, depending on the public IP found). It defaults to `ipv4 or ipv6`.
+- `"ipv6_suffix"` is IPv6 interface identifier suffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw public IPv6 address obtained is used in record updating.
+
+## Domain setup
+
+[![Netlify Website](../readme/netlify.png)](https://www.netlify.com)
+
+1. Login to your Netlify account at [https://app.netlify.com/](https://app.netlify.com/)
+
+2. Navigate to **Site settings** for your site
+
+3. Go to **Domain management** → **DNS zones**
+
+4. Ensure your domain is properly configured as a DNS zone in Netlify
+
+## Token setup
+
+1. Go to **User settings** → **Applications** → **Personal access tokens**
+
+2. Click **New access token**
+
+3. Give the token a descriptive name (e.g., "DDNS Updater")
+
+4. Select the following scopes:
+   - **DNS:read** - Read DNS zones and records
+   - **DNS:edit** - Edit DNS zones and records
+
+5. Click **Generate token**
+
+6. Copy the generated token - this is your `"token"` value
+
+## Testing
+
+1. Go to your Netlify site's DNS management page
+
+2. Check the current DNS record for your domain
+
+3. Run ddns-updater
+
+4. Refresh the Netlify DNS page to verify the update occurred
+
+## Notes
+
+- Netlify's DNS API requires the domain to be configured as a DNS zone in your Netlify account
+- The provider automatically finds the appropriate DNS zone for your domain
+- If a DNS record doesn't exist, it will be created
+- If a DNS record exists with a different IP, it will be deleted and recreated with the new IP
+- The default TTL for created records is 3600 seconds (1 hour)
+- IPv6 is supported if your Netlify account has IPv6 enabled for the DNS zone
--- a/internal/provider/providers/netlify/provider.go
+++ b/internal/provider/providers/netlify/provider.go
@@ -1,0 +1,416 @@
+package netlify
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/netip"
+	"strings"
+
+	"github.com/qdm12/ddns-updater/internal/models"
+	"github.com/qdm12/ddns-updater/internal/provider/constants"
+	"github.com/qdm12/ddns-updater/internal/provider/errors"
+	"github.com/qdm12/ddns-updater/internal/provider/utils"
+	"github.com/qdm12/ddns-updater/pkg/publicip/ipversion"
+)
+
+const defaultTTL = 3600
+
+// Provider is Netlify DNS provider.
+type Provider struct {
+	domain     string
+	owner      string
+	ipVersion  ipversion.IPVersion
+	ipv6Suffix netip.Prefix
+	token      string
+	httpClient *http.Client
+}
+
+// New creates a new Netlify provider.
+func New(data json.RawMessage, domain, owner string,
+	ipVersion ipversion.IPVersion, ipv6Suffix netip.Prefix,
+) (p *Provider, err error) {
+	extraSettings := struct {
+		Token string `json:"token"`
+	}{}
+
+	if err := json.Unmarshal(data, &extraSettings); err != nil {
+		return nil, err
+	}
+
+	p = &Provider{
+		domain:     domain,
+		owner:      owner,
+		ipVersion:  ipVersion,
+		ipv6Suffix: ipv6Suffix,
+		token:      extraSettings.Token,
+		httpClient: &http.Client{},
+	}
+
+	err = p.validateSettings()
+	if err != nil {
+		return nil, fmt.Errorf("validating provider specific settings: %w", err)
+	}
+
+	return p, nil
+}
+
+func (p *Provider) validateSettings() error {
+	if p.token == "" {
+		return fmt.Errorf("%w", errors.ErrTokenNotSet)
+	}
+	return nil
+} // String returns a string representation of provider.
+func (p *Provider) String() string {
+	return utils.ToString(p.domain, p.owner, constants.Netlify, p.ipVersion)
+}
+
+// Domain returns domain of provider.
+func (p *Provider) Domain() string {
+	return p.domain
+}
+
+// Owner returns owner of provider.
+func (p *Provider) Owner() string {
+	return p.owner
+}
+
+// IPVersion returns IP version of provider.
+func (p *Provider) IPVersion() ipversion.IPVersion {
+	return p.ipVersion
+}
+
+// IPv6Suffix returns IPv6 suffix of provider.
+func (p *Provider) IPv6Suffix() netip.Prefix {
+	return p.ipv6Suffix
+}
+
+// BuildDomainName builds the full domain name.
+func (p *Provider) BuildDomainName() string {
+	return utils.BuildDomainName(p.owner, p.domain)
+}
+
+// Proxied returns whether the provider is proxied.
+func (p *Provider) Proxied() bool {
+	return false
+}
+
+// HTML returns HTML representation of provider.
+func (p *Provider) HTML() models.HTMLRow {
+	return models.HTMLRow{
+		Domain:    fmt.Sprintf("<a href=\"http://%s\">%s</a>", p.BuildDomainName(), p.BuildDomainName()),
+		Owner:     p.Owner(),
+		Provider:  "<a href=\"https://www.netlify.com\">Netlify</a>",
+		IPVersion: p.ipVersion.String(),
+	}
+}
+
+// Update updates DNS record.
+func (p *Provider) Update(ctx context.Context, _ *http.Client, ip netip.Addr) (newIP netip.Addr, err error) {
+	// Find the DNS zone for the domain.
+	zone, err := p.findZone(ctx)
+	if err != nil {
+		return netip.Addr{}, fmt.Errorf("failed to find DNS zone: %w", err)
+	}
+
+	// List existing DNS records to find the one to update.
+	records, err := p.listDNSRecords(ctx, zone.ID)
+	if err != nil {
+		return netip.Addr{}, fmt.Errorf("failed to list DNS records: %w", err)
+	}
+
+	// Determine the record type and hostname.
+	recordType := p.getRecordType()
+	hostname := p.getHostname()
+	ipStr := ip.String()
+
+	// Find existing record.
+	var existingRecord *dnsRecord
+	for _, record := range records {
+		if record.Type == recordType && record.Hostname == hostname {
+			existingRecord = &record
+			break
+		}
+	}
+
+	// If record exists and IP is the same, no update needed.
+	if existingRecord != nil && existingRecord.Value == ipStr {
+		return ip, nil
+	}
+
+	// Create or update the record.
+	if existingRecord != nil {
+		err = p.updateDNSRecord(ctx, zone.ID, existingRecord.ID, ipStr)
+		if err != nil {
+			return netip.Addr{}, fmt.Errorf("failed to update DNS record: %w", err)
+		}
+	} else {
+		err = p.createDNSRecord(ctx, zone.ID, hostname, recordType, ipStr)
+		if err != nil {
+			return netip.Addr{}, fmt.Errorf("failed to create DNS record: %w", err)
+		}
+	}
+
+	return ip, nil
+}
+
+func (p *Provider) getRecordType() string {
+	if p.ipVersion == ipversion.IP4 {
+		return "A"
+	}
+	return "AAAA"
+}
+
+func (p *Provider) getHostname() string {
+	if p.owner == "@" {
+		return p.domain
+	}
+	return p.owner + "." + p.domain
+}
+
+// findZone finds the DNS zone for the domain.
+func (p *Provider) findZone(ctx context.Context) (*dnsZone, error) {
+	url := "https://api.netlify.com/api/v1/dns_zones"
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	p.setAuthHeader(req)
+
+	resp, err := p.httpClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("%w: status %d: %s", errors.ErrHTTPStatusNotValid, resp.StatusCode, string(body))
+	}
+
+	var zones []dnsZone
+	if err := json.Unmarshal(body, &zones); err != nil {
+		return nil, err
+	}
+
+	// Find the zone that matches our domain.
+	for _, zone := range zones {
+		if zone.Name == p.domain || strings.HasSuffix(p.domain, "."+zone.Name) {
+			return &zone, nil
+		}
+	}
+
+	return nil, fmt.Errorf("%w: for domain %s", errors.ErrZoneNotFound, p.domain)
+}
+
+// listDNSRecords lists all DNS records for a zone.
+func (p *Provider) listDNSRecords(ctx context.Context, zoneID string) ([]dnsRecord, error) {
+	url := fmt.Sprintf("https://api.netlify.com/api/v1/dns_zones/%s/dns_records", zoneID)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	p.setAuthHeader(req)
+
+	resp, err := p.httpClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("%w: status %d: %s", errors.ErrHTTPStatusNotValid, resp.StatusCode, string(body))
+	}
+
+	var records []dnsRecord
+	if err := json.Unmarshal(body, &records); err != nil {
+		return nil, err
+	}
+
+	return records, nil
+}
+
+// createDNSRecord creates a new DNS record.
+func (p *Provider) createDNSRecord(ctx context.Context, zoneID, hostname, recordType, value string) error {
+	url := fmt.Sprintf("https://api.netlify.com/api/v1/dns_zones/%s/dns_records", zoneID)
+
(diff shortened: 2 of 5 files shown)


docs/netlify.md
```diff
@@ -0,0 +1,76 @@
...
+
+[![Netlify Website](../readme/netlify.png)](https://www.netlify.com)
+
+1. Login to your Netlify account at [https://app.netlify.com/](https://app.netlify.com/)
+
+2. Navigate to **Site settings** for your site
+
+3. Go to **Domain management** → **DNS zones**
+
+4. Ensure your domain is properly configured as a DNS zone in Netlify
```
qdm12, line 41: nit let's use `1.` everywhere so github markdown handles the numbering


```suggestion
1. Navigate to **Site settings** for your site
1. Go to **Domain management** → **DNS zones**
1. Ensure your domain is properly configured as a DNS zone in Netlify
```

docs/netlify.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Token setup
+
```
qdm12, line 44: let's just continue directly to the token setup


```suggestion
```

docs/netlify.md
```diff
@@ -0,0 +1,76 @@
...
+3. Give the token a descriptive name (e.g., "DDNS Updater")
+
+4. Select the following scopes:
+   - **DNS:read** - Read DNS zones and records
+   - **DNS:edit** - Edit DNS zones and records
+
+5. Click **Generate token**
+
+6. Copy the generated token - this is your `"token"` value
```
qdm12, line 57:
```suggestion
1. Click **New access token**
1. Give the token a descriptive name (e.g., "DDNS Updater")
1. Select the following scopes:
   - **DNS:read** - Read DNS zones and records
   - **DNS:edit** - Edit DNS zones and records
1. Click **Generate token** - this is your `"token"` value
```

docs/netlify.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Testing
+
+1. Go to your Netlify site's DNS management page
+
+2. Check the current DNS record for your domain
+
+3. Run ddns-updater
+
+4. Refresh the Netlify DNS page to verify the update occurred
```
qdm12, line 67: I appreciate the effort in documentation, but let's remove this section, I don't think it's really necessary. ddns-updater should log if anything wrong happens.


```suggestion
```

docs/netlify.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Notes
+
+- Netlify's DNS API requires the domain to be configured as a DNS zone in your Netlify account
```
qdm12, line 71: this is already mentioned in the domain setup section


```suggestion
```

docs/netlify.md
```diff
@@ -0,0 +1,76 @@
...
+- The provider automatically finds the appropriate DNS zone for your domain
```
qdm12, line 72: what does that mean?

docs/netlify.md
```diff
@@ -0,0 +1,76 @@
...
+- If a DNS record doesn't exist, it will be created
```
qdm12, line 73: this is expected default behavior (except with a few providers where it's not possible) - let's remove


```suggestion
```

docs/netlify.md
```diff
@@ -0,0 +1,76 @@
...
+- If a DNS record exists with a different IP, it will be deleted and recreated with the new IP
```
qdm12, line 74: that feels like an implementation detail, no need to document it


```suggestion
```

docs/netlify.md
```diff
@@ -0,0 +1,76 @@
...
+- The default TTL for created records is 3600 seconds (1 hour)
```
qdm12, line 75: this should be an optional `"ttl"` parameter, like with other providers. You can mention in defaults to 1hr in the optional parameters section above, and that should also be removed.


```suggestion
```

docs/netlify.md
```diff
@@ -0,0 +1,76 @@
...
+- IPv6 is supported if your Netlify account has IPv6 enabled for the DNS zone
```
qdm12, line 76: Should this be moved to the domain setup section? And when you say "IPv6 enabled", what does that mean exactly? Is there a setting/tickbox on netlify's UI?

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+const defaultTTL = 3600
+
+// Provider is Netlify DNS provider.
+type Provider struct {
+	domain     string
+	owner      string
+	ipVersion  ipversion.IPVersion
+	ipv6Suffix netip.Prefix
+	token      string
+	httpClient *http.Client
```
qdm12, line 29: you should be using the httpClient injected to the Update function instead


```suggestion
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+
+	return p, nil
+}
+
+func (p *Provider) validateSettings() error {
+	if p.token == "" {
+		return fmt.Errorf("%w", errors.ErrTokenNotSet)
+	}
+	return nil
+} // String returns a string representation of provider.
```
qdm12, line 66:
```suggestion
}

// String returns a string representation of provider.
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	token      string
+	httpClient *http.Client
+}
+
+// New creates a new Netlify provider.
+func New(data json.RawMessage, domain, owner string,
+	ipVersion ipversion.IPVersion, ipv6Suffix netip.Prefix,
+) (p *Provider, err error) {
+	extraSettings := struct {
+		Token string `json:"token"`
```
qdm12, line 37: add ttl

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	if p.token == "" {
+		return fmt.Errorf("%w", errors.ErrTokenNotSet)
+	}
+	return nil
+} // String returns a string representation of provider.
+func (p *Provider) String() string {
+	return utils.ToString(p.domain, p.owner, constants.Netlify, p.ipVersion)
+}
+
+// Domain returns domain of provider.
```
qdm12, line 71: let's remove the few comments on common exported methods. These method names are self explicit and don't need comments.

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+		Domain:    fmt.Sprintf("<a href=\"http://%s\">%s</a>", p.BuildDomainName(), p.BuildDomainName()),
+		Owner:     p.Owner(),
+		Provider:  "<a href=\"https://www.netlify.com\">Netlify</a>",
+		IPVersion: p.ipVersion.String(),
+	}
+}
+
+// Update updates DNS record.
+func (p *Provider) Update(ctx context.Context, _ *http.Client, ip netip.Addr) (newIP netip.Addr, err error) {
+	// Find the DNS zone for the domain.
```
qdm12, line 113: unneeded comment


```suggestion
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	return models.HTMLRow{
+		Domain:    fmt.Sprintf("<a href=\"http://%s\">%s</a>", p.BuildDomainName(), p.BuildDomainName()),
+		Owner:     p.Owner(),
+		Provider:  "<a href=\"https://www.netlify.com\">Netlify</a>",
+		IPVersion: p.ipVersion.String(),
+	}
+}
+
+// Update updates DNS record.
+func (p *Provider) Update(ctx context.Context, _ *http.Client, ip netip.Addr) (newIP netip.Addr, err error) {
```
qdm12, line 112: change this to use the passed http client

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	// Find the DNS zone for the domain.
+	zone, err := p.findZone(ctx)
+	if err != nil {
+		return netip.Addr{}, fmt.Errorf("failed to find DNS zone: %w", err)
```
qdm12, line 116: change error wrappings (here and in other places) to be in the form "doing xyz: %w" otherwise it repeats a lot with "failed to do xyz: failed to do abc: ..." when having a full error message

```suggestion
		return netip.Addr{}, fmt.Errorf("finding DNS zone: %w", err)
```

Please review all uses of `fmt.Errorf("failed to`

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	// List existing DNS records to find the one to update.
```
qdm12, line 119: unneeded comment


```suggestion
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+		Owner:     p.Owner(),
+		Provider:  "<a href=\"https://www.netlify.com\">Netlify</a>",
+		IPVersion: p.ipVersion.String(),
+	}
+}
+
+// Update updates DNS record.
+func (p *Provider) Update(ctx context.Context, _ *http.Client, ip netip.Addr) (newIP netip.Addr, err error) {
+	// Find the DNS zone for the domain.
+	zone, err := p.findZone(ctx)
```
qdm12, line 114: only zone.ID is used from findZone, so change it to return the zoneID only

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+		return netip.Addr{}, fmt.Errorf("failed to find DNS zone: %w", err)
+	}
+
+	// List existing DNS records to find the one to update.
+	records, err := p.listDNSRecords(ctx, zone.ID)
+	if err != nil {
+		return netip.Addr{}, fmt.Errorf("failed to list DNS records: %w", err)
+	}
+
+	// Determine the record type and hostname.
```
qdm12, line 125: unneeded comment


```suggestion
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	recordType := p.getRecordType()
+	hostname := p.getHostname()
+	ipStr := ip.String()
+
+	// Find existing record.
```
qdm12, line 130: unneeded comment (which is good, your code is clear 👍)


```suggestion
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	var existingRecord *dnsRecord
+	for _, record := range records {
+		if record.Type == recordType && record.Hostname == hostname {
+			existingRecord = &record
+			break
+		}
+	}
+
+	// If record exists and IP is the same, no update needed.
```
qdm12, line 139:
```suggestion
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+			return netip.Addr{}, fmt.Errorf("failed to update DNS record: %w", err)
+		}
+	} else {
+		err = p.createDNSRecord(ctx, zone.ID, hostname, recordType, ipStr)
+		if err != nil {
+			return netip.Addr{}, fmt.Errorf("failed to create DNS record: %w", err)
+		}
+	}
+
+	return ip, nil
```
qdm12, line 157: nit just reducing indentation + having early returns + having create-record as a particular case in if condition


```suggestion
	if existingRecord == nil {
		err = p.createDNSRecord(ctx, zone.ID, hostname, recordType, ipStr)
		if err != nil {
			return netip.Addr{}, fmt.Errorf("failed to create DNS record: %w", err)
		}
		return ip, nil
	}
	
	err = p.updateDNSRecord(ctx, zone.ID, existingRecord.ID, ipStr)
	if err != nil {
		return netip.Addr{}, fmt.Errorf("failed to update DNS record: %w", err)
	}
	return ip, nil
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+}
+
+func (p *Provider) getRecordType() string {
```
qdm12, line 160: this is used in only one place, just inline the function body at that single place (like with other providers). This simple function body doesn't require a function, it's easier to have it inline for readers instead of having to look elsewhere

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	if p.ipVersion == ipversion.IP4 {
+		return "A"
+	}
+	return "AAAA"
+}
+
+func (p *Provider) getHostname() string {
```
qdm12, line 167: same comment as getRecordType: inline this where needed

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	if p.owner == "@" {
+		return p.domain
+	}
+	return p.owner + "." + p.domain
+}
+
+// findZone finds the DNS zone for the domain.
+func (p *Provider) findZone(ctx context.Context) (*dnsZone, error) {
+	url := "https://api.netlify.com/api/v1/dns_zones"
```
qdm12, line 176: nit use constant whenever possible


```suggestion
	const url = "https://api.netlify.com/api/v1/dns_zones"
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
```
qdm12, line 196: nit check the status before reading the entire body


same comment for `listDNSRecords`, `getDNSRecord`

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+		return nil, fmt.Errorf("%w: status %d: %s", errors.ErrHTTPStatusNotValid, resp.StatusCode, string(body))
+	}
+
+	var zones []dnsZone
+	if err := json.Unmarshal(body, &zones); err != nil {
```
qdm12, line 201: nit you could use a `json.NewDecoder(resp.Body)` to stream decode the json body, instead of having to read it entirely and then parse it.

same comment for `listDNSRecords`, `getDNSRecord`

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	return nil, fmt.Errorf("%w: for domain %s", errors.ErrZoneNotFound, p.domain)
+}
+
+// listDNSRecords lists all DNS records for a zone.
+func (p *Provider) listDNSRecords(ctx context.Context, zoneID string) ([]dnsRecord, error) {
+	url := fmt.Sprintf("https://api.netlify.com/api/v1/dns_zones/%s/dns_records", zoneID)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
```
qdm12, line 219: nit let's use `request` and `response` instead of `req` and `resp`, favor explicit variable names (unless it's a very-very-common name like `i` for the index in a loop for example)

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+
+	var records []dnsRecord
+	if err := json.Unmarshal(body, &records); err != nil {
+		return nil, err
+	}
+
+	return records, nil
+}
+
+// createDNSRecord creates a new DNS record.
```
qdm12, line 249:
```suggestion
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	for _, zone := range zones {
+		if zone.Name == p.domain || strings.HasSuffix(p.domain, "."+zone.Name) {
+			return &zone, nil
+		}
+	}
+
+	return nil, fmt.Errorf("%w: for domain %s", errors.ErrZoneNotFound, p.domain)
+}
+
+// listDNSRecords lists all DNS records for a zone.
```
qdm12, line 215:
```suggestion
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+
+// createDNSRecord creates a new DNS record.
+func (p *Provider) createDNSRecord(ctx context.Context, zoneID, hostname, recordType, value string) error {
+	url := fmt.Sprintf("https://api.netlify.com/api/v1/dns_zones/%s/dns_records", zoneID)
+
+	record := dnsRecordCreate{
+		Type:     recordType,
+		Hostname: hostname,
+		Value:    value,
+		TTL:      defaultTTL, // Default TTL
```
qdm12, line 257: make this an optional parameter

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+		return err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewReader(body))
+	if err != nil {
+		return err
+	}
+
+	p.setAuthHeader(req)
+	req.Header.Set("Content-Type", "application/json")
```
qdm12, line 271: you can use the `headers` package I think with `headers.SetContentType`

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+
+	resp, err := p.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusCreated {
+		body, _ := io.ReadAll(resp.Body)
```
qdm12, line 280: use `utils.BodyToSingleLine`

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+		return fmt.Errorf("%w: status %d: %s", errors.ErrHTTPStatusNotValid, resp.StatusCode, string(body))
+	}
+
+	return nil
+}
+
+// updateDNSRecord updates an existing DNS record.
```
qdm12, line 287:
```suggestion
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+func (p *Provider) updateDNSRecord(ctx context.Context, zoneID, recordID, value string) error {
+	// Netlify API doesn't have a direct update endpoint for DNS records.
+	// We need to delete the existing record and create a new one.
+
+	// First, get the existing record details.
```
qdm12, line 292: Last suggestion on removing unneeded comments; please review all your method and inline comments, most of them are unneeded, so please remove them.

```suggestion
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	UpdatedAt            string      `json:"updated_at"`
+	Records              []dnsRecord `json:"records"`
+	DNSServers           []string    `json:"dns_servers"`
+	AccountID            string      `json:"account_id"`
+	SiteID               string      `json:"site_id"`
+	AccountSlug          string      `json:"account_slug"`
+	AccountName          string      `json:"account_name"`
+	Domain               string      `json:"domain"`
+	IPv6Enabled          bool        `json:"ipv6_enabled"`
+	Dedicated            bool        `json:"dedicated"`
```
qdm12, line 387: only `Name` and `ID` are used in this struct, so remove the other fields


```suggestion
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	CreatedAt            string      `json:"created_at"`
+	UpdatedAt            string      `json:"updated_at"`
+	Records              []dnsRecord `json:"records"`
+	DNSServers           []string    `json:"dns_servers"`
+	AccountID            string      `json:"account_id"`
+	SiteID               string      `json:"site_id"`
+	AccountSlug          string      `json:"account_slug"`
+	AccountName          string      `json:"account_name"`
+	Domain               string      `json:"domain"`
+	IPv6Enabled          bool        `json:"ipv6_enabled"`
```
qdm12, line 386: that field should be used in case the user tries to set an ipv6 address and this is false, to return a meaningful error

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+// updateDNSRecord updates an existing DNS record.
+func (p *Provider) updateDNSRecord(ctx context.Context, zoneID, recordID, value string) error {
+	// Netlify API doesn't have a direct update endpoint for DNS records.
+	// We need to delete the existing record and create a new one.
+
+	// First, get the existing record details.
+	record, err := p.getDNSRecord(ctx, zoneID, recordID)
+	if err != nil {
+		return fmt.Errorf("failed to get existing DNS record: %w", err)
+	}
```
qdm12, line 296: do we need to get the record information only to have `record.Hostname`? Can't we build it locally without doing a request?

It feels like getDNSRecord could be removed

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+type dnsRecordCreate struct {
+	Type     string `json:"type"`
+	Hostname string `json:"hostname"`
+	Value    string `json:"value"`
+	TTL      int64  `json:"ttl"`
+	Priority int64  `json:"priority,omitempty"`
+	Weight   int64  `json:"weight,omitempty"`
+	Port     int64  `json:"port,omitempty"`
+	Flag     int64  `json:"flag,omitempty"`
+	Tag      string `json:"tag,omitempty"`
```
qdm12, line 415: unused so remove them


```suggestion
```

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	Tag       string `json:"tag"`
+	Managed   bool   `json:"managed"`
+}
+
+// DNS record creation structure.
+type dnsRecordCreate struct {
+	Type     string `json:"type"`
+	Hostname string `json:"hostname"`
+	Value    string `json:"value"`
+	TTL      int64  `json:"ttl"`
```
qdm12, line 410: what happens if we leave this empty


```suggestion
	TTL      int64  `json:"ttl,omitempty"`
```

does netlify set a default ttl on their own? That might be more appropriate than having a hardcoded default ttl in ddns-updater

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	Priority  int64  `json:"priority"`
+	DNSZoneID string `json:"dns_zone_id"`
+	SiteID    string `json:"site_id"`
+	Flag      int64  `json:"flag"`
+	Tag       string `json:"tag"`
+	Managed   bool   `json:"managed"`
+}
+
+// DNS record creation structure.
+type dnsRecordCreate struct {
```
qdm12, line 406: since this is used in only one place, I would recommend declaring this type where it's used inside the method instead of at global scope.

internal/provider/providers/netlify/provider.go
```diff
@@ -0,0 +1,416 @@
...
+	SiteID               string      `json:"site_id"`
+	AccountSlug          string      `json:"account_slug"`
+	AccountName          string      `json:"account_name"`
+	Domain               string      `json:"domain"`
+	IPv6Enabled          bool        `json:"ipv6_enabled"`
+	Dedicated            bool        `json:"dedicated"`
+}
+
+// DNS record structure.
+type dnsRecord struct {
```
qdm12, line 391: since this **will** be used in only one place, I would recommend declaring this type where it's used inside the method instead of at global scope. You can also remove all fields except `ID` as far as I recall, other fields are not used.

qdm12 (changes requested):

PR #1065, opened 2025-12-27
Added Vercel Support

williambout:
This PR adds support for Vercel Domains

### DIFF
--- a/docs/vercel.md
+++ b/docs/vercel.md
@@ -1,0 +1,56 @@
+# Vercel
+
+## Configuration
+
+### Example
+
+```json
+{
+  "settings": [
+    {
+      "provider": "vercel",
+      "domain": "domain.com",
+      "token": "yourtoken",
+      "ip_version": "ipv4",
+      "ipv6_suffix": ""
+    }
+  ]
+}
+```
+
+### Compulsory parameters
+
+- `"domain"` is the domain to update. It can be `example.com` (root domain), `sub.example.com` (subdomain of `example.com`) or `*.example.com` for the wildcard.
+- `"token"` is your Vercel API token. You can create one at [Vercel Account Settings > Tokens](https://vercel.com/account/tokens).
+
+### Optional parameters
+
+- `"team_id"` is your Vercel team ID. Required if the domain belongs to a team rather than your personal account. You can find this in your team settings.
+- `"ttl"` is the TTL in seconds for the DNS record. Defaults to `60` if not specified.
+- `"ip_version"` can be `ipv4` (A records), or `ipv6` (AAAA records) or `ipv4 or ipv6` (update one of the two, depending on the public ip found). It defaults to `ipv4 or ipv6`.
+- `"ipv6_suffix"` is the IPv6 interface identifier suffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw public IPv6 address obtained is used in the record updating.
+
+## Domain setup
+
+1. Ensure your domain is added to Vercel. You can do this via the [Vercel Dashboard](https://vercel.com/dashboard) under your project's Domain settings, or via the API.
+2. Your domain must be using Vercel's nameservers or have DNS management delegated to Vercel for this provider to work.
+3. Create an API token with appropriate permissions at [Vercel Account Settings > Tokens](https://vercel.com/account/tokens).
+4. If the domain belongs to a team, make sure to include the `team_id` parameter in your configuration.
+
+## Example with team
+
+```json
+{
+  "settings": [
+    {
+      "provider": "vercel",
+      "domain": "domain.com",
+      "token": "yourtoken",
+      "team_id": "team_xxxxxxxxxx",
+      "ttl": 300,
+      "ip_version": "ipv4"
+    }
+  ]
+}
+```
+
--- a/internal/provider/providers/vercel/provider.go
+++ b/internal/provider/providers/vercel/provider.go
@@ -1,0 +1,305 @@
+package vercel
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	stderrors "errors"
+	"fmt"
+	"net/http"
+	"net/netip"
+	"net/url"
+
+	"github.com/qdm12/ddns-updater/internal/models"
+	"github.com/qdm12/ddns-updater/internal/provider/constants"
+	"github.com/qdm12/ddns-updater/internal/provider/errors"
+	"github.com/qdm12/ddns-updater/internal/provider/headers"
+	"github.com/qdm12/ddns-updater/internal/provider/utils"
+	"github.com/qdm12/ddns-updater/pkg/publicip/ipversion"
+)
+
+type Provider struct {
+	domain     string
+	owner      string
+	ipVersion  ipversion.IPVersion
+	ipv6Suffix netip.Prefix
+	token      string
+	teamID     string
+	ttl        uint32
+}
+
+func New(data json.RawMessage, domain, owner string,
+	ipVersion ipversion.IPVersion, ipv6Suffix netip.Prefix) (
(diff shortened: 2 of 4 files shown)


docs/vercel.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Compulsory parameters
+
+- `"domain"` is the domain to update. It can be `example.com` (root domain), `sub.example.com` (subdomain of `example.com`) or `*.example.com` for the wildcard.
+- `"token"` is your Vercel API token. You can create one at [Vercel Account Settings > Tokens](https://vercel.com/account/tokens).
+
+### Optional parameters
+
+- `"team_id"` is your Vercel team ID. Required if the domain belongs to a team rather than your personal account. You can find this in your team settings.
+- `"ttl"` is the TTL in seconds for the DNS record. Defaults to `60` if not specified.
```
qdm12, line 29: 60 is quite short, is there a specific reason for this? usually it's 300 or 600

docs/vercel.md
```diff
@@ -0,0 +1,56 @@
...
+- `"ip_version"` can be `ipv4` (A records), or `ipv6` (AAAA records) or `ipv4 or ipv6` (update one of the two, depending on the public ip found). It defaults to `ipv4 or ipv6`.
+- `"ipv6_suffix"` is the IPv6 interface identifier suffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw public IPv6 address obtained is used in the record updating.
+
+## Domain setup
+
+1. Ensure your domain is added to Vercel. You can do this via the [Vercel Dashboard](https://vercel.com/dashboard) under your project's Domain settings, or via the API.
+2. Your domain must be using Vercel's nameservers or have DNS management delegated to Vercel for this provider to work.
+3. Create an API token with appropriate permissions at [Vercel Account Settings > Tokens](https://vercel.com/account/tokens).
+4. If the domain belongs to a team, make sure to include the `team_id` parameter in your configuration.
```
qdm12, line 38: let github markdown handle the numbering 😉 

```suggestion
1. Ensure your domain is added to Vercel. You can do this via the [Vercel Dashboard](https://vercel.com/dashboard) under your project's Domain settings, or via the API.
1. Your domain must be using Vercel's nameservers or have DNS management delegated to Vercel for this provider to work.
1. Create an API token with appropriate permissions at [Vercel Account Settings > Tokens](https://vercel.com/account/tokens).
1. If the domain belongs to a team, make sure to include the `team_id` parameter in your configuration.
```

docs/vercel.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Example with team
+
+```json
+{
```
qdm12, line 43: perhaps move this right below the original example?

internal/provider/providers/vercel/provider.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+func (p *Provider) setHeaders(request *http.Request) {
+	headers.SetUserAgent(request)
+	headers.SetAccept(request, "application/json")
+	headers.SetContentType(request, "application/json")
+	headers.SetAuthBearer(request, p.token)
+}
+
+func (p *Provider) makeURL(path string) string {
```
qdm12: nit we could have this as a pure function and pass in the `teamID` as extra argument i.e.


```suggestion
func (p *Provider) makeURL(teamID, path string) string {
```

just to make it clear that function is here for the teamID 😉

internal/provider/providers/vercel/provider.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	// Find the matching record by name and type
+	targetName := p.owner
+	if targetName == "@" {
+		targetName = ""
+	}
+
+	for _, r := range result.Records {
+		if r.Name == targetName && r.Type == recordType {
```
qdm12, line 183: nit just for extra safety:


```suggestion
		if r.Name == targetName && strings.EqualFold(r.Type, recordType) {
```

internal/provider/providers/vercel/provider.go
```diff
@@ -0,0 +1,305 @@
...
+
+type dnsRecord struct {
+	ID    string `json:"id"`
+	Name  string `json:"name"`
+	Type  string `json:"type"`
+	Value string `json:"value"`
+	TTL   uint32 `json:"ttl"`
+}
+
+func (p *Provider) getRecord(ctx context.Context, client *http.Client, recordType string) (
```
qdm12, line 145: nit since only the id and value fields are used from the returned record, let's change this to return `id, value string, err error`? Then you can inline the `dnsRecord` type definition in `var result` below as well.

Or, if you feel it's appropriate to return the whole record, let's return a `dnsRecord` (not `*dnsRecord`) since there is no need for a pointer to dnsRecord. You can just return `dnsRecord{}, err` when encountering an error

internal/provider/providers/vercel/provider.go
```diff
@@ -0,0 +1,305 @@
...
+	recordType := constants.A
+	if ip.Is6() {
+		recordType = constants.AAAA
+	}
+
+	u := p.makeURL("/v4/domains/" + p.domain + "/records")
+
+	name := p.owner
+	if name == "@" {
+		name = ""
```
qdm12, line 201: so their root domain rr has name `""` ewww strange!

internal/provider/providers/vercel/provider.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	requestData := struct {
+		Name    string `json:"name"`
+		Type    string `json:"type"`
+		Value   string `json:"value"`
+		TTL     uint32 `json:"ttl,omitempty"`
```
qdm12, line 208: if this is omitted (aka set to 0 here), what happens? Does vercel set a default ttl? This would seem more appropriate instead of having a default ttl in ddns-updater

internal/provider/providers/vercel/provider.go
```diff
@@ -0,0 +1,305 @@
...
+
+	if response.StatusCode != http.StatusOK && response.StatusCode != http.StatusNoContent {
+		return fmt.Errorf("%w: %d: %s",
+			errors.ErrHTTPStatusNotValid, response.StatusCode, utils.BodyToSingleLine(response.Body))
+	}
+
+	return nil
+}
+
+func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Addr) (newIP netip.Addr, err error) {
```
qdm12, line 268: nit put this above all unexported methods that are part of its body. You could also eventually move each function to its own file, if you want, i.e. create.go, delete.go etc.

internal/provider/providers/vercel/provider.go
```diff
@@ -0,0 +1,305 @@
...
+	recordType := constants.A
+	if ip.Is6() {
+		recordType = constants.AAAA
+	}
+
+	record, err := p.getRecord(ctx, client, recordType)
+	switch {
+	case stderrors.Is(err, errors.ErrReceivedNoResult):
+		// Record doesn't exist, create it
```
qdm12, line 277: nit unneeded comment imo


```suggestion
```

internal/provider/providers/vercel/provider.go
```diff
@@ -0,0 +1,305 @@
...
+	case err != nil:
+		return netip.Addr{}, fmt.Errorf("getting record: %w", err)
+	}
+
+	// Check if IP is already up to date
+	if record.Value == ip.String() {
+		return ip, nil
+	}
+
+	// Delete the existing record and create a new one with the updated IP
```
qdm12, line 292: Is it not possible to update the record directly? Does it have to be deleted and re-created?

qdm12 (commented):
This being my 3rd new provider-PR of my Saturday morning, congrats this made me very happy to (finally) see good code 😄 👍 💯

PR #1093, opened 2026-02-09
perf(persistence): replace O(n) linear scans with O(1) index lookups

### LINKED ISSUE #1092: perf: replace O(n) linear scans with O(1) index lookups in JSON database
## Problem

`StoreNewIP()` and `GetEvents()` in `internal/persistence/json/queries.go` both perform O(n) linear scans over all records to find the matching domain/owner pair:

```go
for i, record := range db.data.Records {
    if record.Domain == domain && record.Owner == owner {
        targetIndex = i
        break
    }
}
```

Every DNS update cycle calls `StoreNewIP()`, so this scan runs frequently. For deployments with many records, this is unnecessarily expensive.

## Proposed Solution

Maintain an in-memory index map (`map[string]int`) keyed by `domain + "\x00" + owner` that maps to the record's index in the slice. This gives O(1) lookups instead of O(n) scans.

The index should be:
- Built at database load time in `NewDatabase()`
- Updated when new records are appended in `StoreNewIP()`
- Used in both `StoreNewIP()` and `GetEvents()`

## Impact

Low risk, high value for users with many DNS records. The change is internal to the persistence layer with no API changes.

fredericrous:
## Summary

- Adds an in-memory index map (`map[string]int`) to the JSON database, keyed by `"domain\x00owner"`, mapping to the record's slice index
- Replaces linear scans in `StoreNewIP()` and `GetEvents()` with O(1) map lookups
- Index is built at database load time and kept in sync when new records are appended

Fixes #1092

## Test plan

- [x] `go test ./...` — all existing tests pass
- [ ] Manual: verify IP updates persist correctly with multiple records
- [ ] Manual: verify database loads correctly from existing `updates.json`

### DIFF
--- a/internal/persistence/json/database.go
+++ b/internal/persistence/json/database.go
@@ -16,6 +16,21 @@ type Database struct {
 	data     dataModel
 	filepath string
 	mutex    sync.RWMutex
+	// index maps "domain\x00owner" to the record index in data.Records
+	// for O(1) lookups instead of linear scans.
+	index map[string]int
+}
+
+func recordKey(domain, owner string) string {
+	return domain + "\x00" + owner
+}
+
+func buildIndex(records []record) map[string]int {
+	index := make(map[string]int, len(records))
+	for i, r := range records {
+		index[recordKey(r.Domain, r.Owner)] = i
+	}
+	return index
 }
 
 func (db *Database) Close() error {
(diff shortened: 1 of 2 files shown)


fredericrous:
Closing — with ~52 records the linear scan is negligible; the real bottleneck is elsewhere.

PR #1095, opened 2026-02-09
perf: reuse HTTP client transport instead of cloning it

### LINKED ISSUE #1094: perf: makeLogClient clones transport, defeating HTTP connection pooling
## Problem

`makeLogClient()` in `internal/update/logclient.go` creates a **new** `http.Client` with a `transport.Clone()`. This means the logging client has its own separate connection pool, so TCP+TLS connections established by the original client are never reused — and vice versa.

Since only the logging client is used for provider updates (`update.go:27`), the original client's pool is wasted, and the logging client rebuilds connections from scratch.

More importantly, the function signature and approach are unnecessarily complex. All that's needed is to wrap the original transport with a `loggingRoundTripper` — no new client or cloned transport required.

## Proposed Solution

Instead of cloning the transport and creating a new client, wrap the original client's transport with `loggingRoundTripper` directly. This preserves the original connection pool and keep-alive behavior.

## Impact

Eliminates redundant TCP+TLS handshakes on every update cycle. With 52 records hitting various DNS providers, this saves real network round trips.

fredericrous:
## Summary

- `makeLogClient()` was creating a new `http.Client` with `transport.Clone()`, giving it a separate connection pool
- TCP+TLS connections were never reused between update cycles — every cycle paid for new handshakes
- Now wraps the original client's transport with `loggingRoundTripper` directly, preserving the connection pool and keep-alive behavior
- Net effect: **-12 lines**, simpler code, fewer allocations, real network savings

Fixes #1094

## Test plan

- [x] `go test ./...` — all tests pass including `Test_LogClient`
- [ ] Manual: verify debug logging still works when enabled
- [ ] Manual: verify DNS updates succeed with the shared transport

### DIFF
--- a/internal/update/logclient.go
+++ b/internal/update/logclient.go
@@ -2,7 +2,6 @@ package update
 
 import (
 	"bytes"
-	"fmt"
 	"io"
 	"net/http"
 	"strings"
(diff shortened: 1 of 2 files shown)


qdm12 (approved):
Sure, I'm not sure why the transport was clone anyway

PR #1097, opened 2026-02-09
perf: skip HTTP body reading when debug logging is disabled

### LINKED ISSUE #1096: perf: skip HTTP body reading when debug logging is disabled
## Problem

The `loggingRoundTripper` in `internal/update/logclient.go` unconditionally reads the entire HTTP request and response body via `io.ReadAll()` on every round trip, even when the log level is Info (the default).

The bodies are fully read into memory, converted to strings, and concatenated — then passed to `Debug()` which silently discards them because the level check happens inside the logger, **after** all the expensive work is done.

With 52 records, this means dozens of unnecessary full body reads per update cycle.

## Proposed Solution

When debug logging is not enabled, skip wrapping the HTTP transport with `loggingRoundTripper` entirely. This avoids all body reading, string allocation, and concatenation overhead.

## Impact

Eliminates unnecessary CPU and memory usage on every HTTP call in the default (non-debug) configuration.

fredericrous:
## Summary

- `loggingRoundTripper` was reading entire HTTP request/response bodies via `io.ReadAll()` on **every** round trip, even at the default Info log level
- The expensive work (body read, string conversion, concatenation) happened **before** `Debug()` was called, so the logger's internal level check couldn't prevent it
- Now `makeLogClient` skips wrapping the transport entirely when debug logging is not enabled — zero overhead in the common case
- Stacked on #1095

Fixes #1096

## Test plan

- [x] `go test ./...` — all tests pass
- [ ] Manual: verify `LOG_LEVEL=debug` still produces HTTP debug output
- [ ] Manual: verify default log level no longer reads bodies (observable via reduced CPU/memory)


internal/update/update.go
```diff
@@ -22,9 +22,9 @@ type Updater struct {
 }
 
 func NewUpdater(db Database, client *http.Client, shoutrrrClient ShoutrrrClient,
-	logger DebugLogger, timeNow func() time.Time,
+	logger DebugLogger, timeNow func() time.Time, debugEnabled bool,
 ) *Updater {
-	client = makeLogClient(client, logger)
+	client = makeLogClient(client, logger, debugEnabled)
```
qdm12: let's just re-assign client if debugEnabled is true here, and leave makeLogClient as it is

qdm12 (commented):
Rebase on master/merge on master

cmd/ddns-updater/main.go
```diff
@@ -206,7 +207,8 @@ func _main(ctx context.Context, reader *reader.Reader, args []string, logger log
 	hioClient := healthchecksio.New(client, config.Health.HealthchecksioBaseURL,
 		*config.Health.HealthchecksioUUID)
 
-	updater := update.NewUpdater(db, client, shoutrrrClient, logger, timeNow)
+	debugEnabled := strings.EqualFold(config.Logger.Level, "DEBUG")
```
qdm12, line 210: (and import the qdm12/log package in main.go)

```suggestion
	debugEnabled := config.Logger.Level == log.LevelDebug.String()
```

qdm12 (approved):
Noice 👍  Just minor enhancement

PR #1098, opened 2026-02-11
docs(porkbun): move ttl to optional parameters section

nbeal:
Porkbun requires the host parameter to work properly and the docs did not have that at all. Added with some examples and moved TTL to the optional parameters.


docs/porkbun.md
```diff
@@ -22,14 +23,19 @@
...
+  * `"*,@"` to update both wildcard and root records
 - `"api_key"`
 - `"secret_api_key"`
-- `"ttl"` optional integer value corresponding to a number of seconds
 
 ### Optional parameters
 
 - `"ip_version"` can be `ipv4` (A records), or `ipv6` (AAAA records) or `ipv4 or ipv6` (update one of the two, depending on the public ip found). It defaults to `ipv4 or ipv6`.
 - `"ipv6_suffix"` is the IPv6 interface identifier suffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw public IPv6 address obtained is used in the record updating.
+- - `"ttl"` optional integer value corresponding to a number of seconds
```
qdm12, line 38: Nice

docs/porkbun.md
```diff
@@ -22,14 +23,19 @@
 ### Compulsory Parameters
 
 - `"domain"` is the domain to update. It can be `example.com` (root domain), `sub.example.com` (subdomain of `example.com`) or `*.example.com` for the wildcard.
+- `"host"` is the DNS host to update:
+  * `"@"` for the root domain (example.com)
+  * `"sub"` for sub.example.com
+  * `"*"` for the wildcard (*.example.com)
+  * `"*,@"` to update both wildcard and root records
```
qdm12, line 30: This is now integrated in the `"domain"` key, no need to use this one. That was in previous versions of the program, but now it's clever enough to figure out the host from the domain string. So `*.example.com` means domain is `example.com` and host is `*`. Please revert this addition

qdm12 (changes requested):

PR #1046, opened 2025-10-26
feat(provider): add hetznercloud

likt0r:
(content replaced by `qdm12`)

Build the image with:

```
docker build -t qmcgaw/ddns-updater:hetzner https://github.com/likt0r/ddns-updater.git#add-provider-hetznernetworking
```

And then run with image `qmcgaw/ddns-updater:hetzner`

[Configuration Documentation](https://github.com/likt0r/ddns-updater/blob/add-provider-hetznernetworking/docs/hetznercloud.md)

Original code by @likt0r , reworked by @qdm12 and completed with bits from #1111 @legend813

PR #1111, opened 2026-04-08
feat(provider): Hetzner Cloud API Provider hinzufügen

legend813:
(replaced by qdm12)

See #1046 instead, this PR is to be closed.

### DIFF
--- a/internal/provider/providers/hetznercloud/getrrset.go
+++ b/internal/provider/providers/hetznercloud/getrrset.go
@@ -1,0 +1,78 @@
+package hetznercloud
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/netip"
+	"net/url"
+
+	"github.com/qdm12/ddns-updater/internal/provider/constants"
+	"github.com/qdm12/ddns-updater/internal/provider/errors"
+	"github.com/qdm12/ddns-updater/internal/provider/utils"
+)
+
+// getRRSet liefert die aktuelle IP des passenden RRsets (A oder AAAA) zurück.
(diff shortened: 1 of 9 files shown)


RealFinalCode:
There is already a PR for that API, though it is confusingly named hetznernetworking. See #1046

gregyski:
Hetzner themselves refer to it as Hetzner Cloud API so this is much better naming.

qdm12:
I reworked #1046 and it's now named hetznercloud. I'll migrate whatever's here into #1046

internal/provider/providers/hetznercloud/getrrset.go
```diff
@@ -0,0 +1,78 @@
...
+) {
+	recordType := constants.A
+	if ip.Is6() {
+		recordType = constants.AAAA
+	}
+
+	u := url.URL{
+		Scheme: "https",
+		Host:   "api.hetzner.cloud",
+		Path:   "/v1/zones/" + zoneID + "/rrsets",
```
qdm12, line 30: As far as I understood, `zoneID` can be the zone name, so for `sub.example.com`, it would be `example.com`, so there is no need to set or find a zone identifier, just use the zone name directly.

qdm12 (commented):
@legend813 (and others) can you try the code from #1046 - see the PR description it says how to build the program and how to setup your config.json

and check it works fine for root domains, subdomains and wildcards, if you can please?
Thanks 😉

qdm12:
Closed in favor of #1046 that is now merged!

legend813:
Really looking forward to the next release. Do you already know when you will release the new version? @qdm12

PR #1114, opened 2026-04-10
fix: add proportional jitter to reduce synchronized API calls

jhutchings1:
Hello, I'm a product manager at @Cloudflare who oversees our API platform. Our logs show that we get millions of clients calling our DDNS endpoints every 5 minutes. This is a noticeable increase over our typical baseline, and contributes to performance degradation. We would like to smooth out the API traffic so that the spikes aren't so prominent, and we have identified your project as one to reach out to.

I added some code that introduces proportional jitter to the DDNS update interval, which should help reduce synchronized API calls and improve performance for us and all the other services that you're using.

Let me know if you have any questions on this, happy to take your suggestions on this as well!

### DIFF
--- a/internal/update/service.go
+++ b/internal/update/service.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand"
 	"net/netip"
 	"strconv"
 	"strings"
@@ -413,6 +414,7 @@ func (s *Service) run(ctx context.Context, ready chan<- struct{},
 	for {
 		select {
 		case <-ticker.C:
+			time.Sleep(jitterDuration(s.period))
 			s.updateNecessary(ctx)
 		case <-s.force:
 			s.forceResult <- s.updateNecessary(ctx)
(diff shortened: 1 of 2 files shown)


qdm12:
Hi there!

> Our logs show that we get millions of clients calling our DDNS endpoints every 5 minutes

Which endpoint? The 5 min period only fetches the public ip address of the machine and compares it to a dns resolve of the domain name, it does NOT query a ddns endpoint, unless there is a mismatch.

jhutchings1:
Hey @qdm12 Thanks for your quick response! 

The call spikes we see are around calls to:
* /zones/:zoneid/dns_records/
* /zones/:zoneid/dns_records/:dns_record_id
* /zones/:zoneid/dns_records?name=:name

Our metadata doesn't show us much detail on the identity of the calling software. The user_agent comes back as curl, so we're somewhat guessing as to which services may be contributing. Yours came up in my search of DDNS updaters that might be contributing, but you know your software better than I do, of course! I'm attaching a graph of the traffic pattern in case you're curious.  

Screenshot 2026-04-11 at 7 38 16 AM

qdm12:
The user agent is set to `DDNS-Updater quentin.mcgaw@gmail.com` so if you see mostly `curl` that's likely coming from another place 😉 

https://github.com/qdm12/ddns-updater/blob/02a0ddd7c11f8509884b8437c82bf4233158f86a/internal/provider/providers/cloudflare/provider.go#L152

Just to narrow down the search, since you see spikes every 5 minutes starting from the top of the hour AND you see `curl` user agents, I would guess it's not really a program but more of a cron job + curl, so that'll be a tough one to tweak unfortunately. If it would be 5 min periodically from program start, there wouldn't be much spikes I think, especially not at `:05`, `:10`, ...`, `:00` minutes marks.

jhutchings1:
Thanks @qdm12, we're looking at all options. Appreciate your help here.

PR #1125, opened 2026-04-19
docs(readme): Home Assistant subsection

Blaimi:
Hi guys,

I created a Home Assistant addon for ddns-updater. Would be nice if it could be mentioned in the setup-instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,10 @@ This readme and the [docs/](docs/) directory are **versioned** to match the prog
     - You can update the image with `docker pull ghcr.io/qdm12/ddns-updater`
     - You can set your JSON configuration as a single environment variable line (i.e. `{"settings": [{"provider": "namecheap", ...}]}`), which takes precedence over config.json. Note however that if you don't bind mount the `/updater/data` directory, there won't be a persistent database file `/updater/updates.json` but it will still work.
 
+### Home Assistant
+
+To install ddns-updater on [Home Assistant](https://home-assistant.io/), follow the installation instructions on [ha-ddns-updater](https://github.com/ha-ddns-updater/ha-ddns-updater/).
+
 ## Configuration
 
 Start by having the following content in *config.json*, or in your `CONFIG` environment variable:


README.md
```diff
@@ -194,6 +194,9 @@ This readme and the [docs/](docs/) directory are **versioned** to match the prog
     - You can update the image with `docker pull ghcr.io/qdm12/ddns-updater`
     - You can set your JSON configuration as a single environment variable line (i.e. `{"settings": [{"provider": "namecheap", ...}]}`), which takes precedence over config.json. Note however that if you don't bind mount the `/updater/data` directory, there won't be a persistent database file `/updater/updates.json` but it will still work.
 
+### Home Assistant
```
qdm12, line 198:
```suggestion
### Home Assistant

```

qdm12 (approved):
Thanks! Just add a newline between the text and heading 😉

Blaimi:
May I ask you about your current release workflow? Currently the addon is based on your latest tagged version and has pipelines to check for new versions to poke me via an MR. Relying on `latest` would likely break the addon very often or give me headache with updates. This is because the addon needs a settings schema and translation file which needs to match the settings of this project.

If you want, I can help you integrating https://github.com/semantic-release/semantic-release to this project. It would be the same workflow as at the moment (push to master, build from master, use from latest), but with version numbers (push to master, create tag/release, build from tag, use from tag—or latest) so the users can decide whether they want to pin to a major, minor, or specific version—or always follow the latest ;-) .

To be honest, I only created the addon because of #1046. I couldn't find any active and trustworthy HA addon which provides the new hetzner DNS API so I had to create my own. Instead of building it from scratch I thought it is more useful to build it around an existing and even more useful project which lead me here. And I love it! Not only because it has a webinterface, but especially because of its ability to update even offline devices via IPv6-suffixes (privacy disabled with eui64 of course).

WolfRevo:
@qdm12 @Blaimi I would love to see ddns-updater being released with semver releases on hub.docker.com instead of latest. That would much better fit into todays patch management requirements we have. So I like the idea of Blaimi supporting you here.

qdm12:
> If you want, I can help you integrating https://github.com/semantic-release/semantic-release to this project. It would be the same workflow as at the moment (push to master, build from master, use from latest), but with version numbers (push to master, create tag/release, build from tag, use from tag—or latest) so the users can decide whether they want to pin to a major, minor, or specific version—or always follow the latest

That's already in place. I'm just slow on this project at the moment

qdm12:
@Blaimi @WolfRevo The https://github.com/qdm12/ddns-updater/releases are following semver. But I prefer to keep it manual rather than automated.

PR #1137, opened 2026-05-05
Use correct provider name in Scaleway String()

jcarvalho:
The `String()` function in the new Scaleway provider incorrectly references the `dyn` provider, so this commit updates it to the correct constant.


qdm12 (approved):
Thanks 👍 !

PR #1141, opened 2026-05-15
feat(provider): `hostinger`

### LINKED ISSUE #849: Add support to Hostinger
If possible, please add support do Hostinger provider.
Thank you!

SpecialAro:
## Description

Adds native Hostinger DNS provider support to `ddns-updater`.

This implements the provider requested in #849 using Hostinger’s DNS API. The new provider supports updating `A` and `AAAA` records for a configured domain/owner using a Hostinger API bearer token.

The implementation follows the existing provider structure used in the project and registers Hostinger as a selectable provider through the provider constants and provider factory.

Closes #849.

## Changes

- Add new Hostinger provider implementation
  - Supports IPv4 `A` records
  - Supports IPv6 `AAAA` records
  - Uses Hostinger API bearer token authentication
  - Uses `PUT /api/dns/v1/zones/{domain}` with `overwrite=true`
  - Supports configurable TTL
  - Defaults TTL to `14400` seconds when unset
  - Maps Hostinger API error responses to existing provider errors where possible

- Register `hostinger` as a provider
  - Add `constants.Hostinger`
  - Add Hostinger to provider choices
  - Add Hostinger provider factory case

- Add Hostinger documentation
  - Example JSON configuration
  - Required parameters
  - Optional parameters
  - Basic setup notes

- Update README provider list with Hostinger

## Configuration example

```json
{
  "settings": [
    {
      "provider": "hostinger",
      "domain": "example.com",
      "owner": "@",
      "token": "YOUR_HOSTINGER_API_TOKEN",
      "ttl": 300,
      "ip_version": "ipv4"
    }
  ]
}

### CODE REVIEW

File: docs/hostinger.md [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+    {
+      "provider": "hostinger",
+      "domain": "example.com",
+      "token": "YOUR_HOSTINGER_API_TOKEN",
+      "ttl": 300,
+      "ip_version": "ipv4",
+      "ipv6_suffix": ""
+    }
+  ]
+}
```
qdm12: please conform to the existing documentation format, using a compulsory and optional parameters subsections describing what's what

internal/provider/providers/hostinger/provider.go
```diff
@@ -0,0 +1,218 @@
...
+
+	"github.com/qdm12/ddns-updater/internal/models"
+	"github.com/qdm12/ddns-updater/internal/provider/constants"
+	"github.com/qdm12/ddns-updater/internal/provider/errors"
+	"github.com/qdm12/ddns-updater/internal/provider/headers"
+	"github.com/qdm12/ddns-updater/internal/provider/utils"
+	"github.com/qdm12/ddns-updater/pkg/publicip/ipversion"
+)
+
+const defaultTTL uint32 = 14400
```
qdm12: nit define this inline right above

```
if extraSettings.TTL == 0 {
```

internal/provider/providers/hostinger/provider.go
```diff
@@ -0,0 +1,218 @@
...
+func validateSettings(domain, token string) (err error) {
+	err = utils.CheckDomain(domain)
+	if err != nil {
+		return fmt.Errorf("%w: %w", errors.ErrDomainNotValid, err)
+	}
+
+	switch {
+	case token == "":
+		return fmt.Errorf("%w", errors.ErrTokenNotSet)
+	}
```
qdm12: or the ci will complain

```suggestion
	if token == "" {
		return fmt.Errorf("%w", errors.ErrTokenNotSet)
	}
```

internal/provider/providers/hostinger/provider.go
```diff
@@ -0,0 +1,218 @@
...
+		return netip.Addr{}, handleAPIError(response)
+	}
+
+	return ip, nil
+}
+
+func handleAPIError(response *http.Response) error {
+	apiError := apiError{}
+	if err := json.NewDecoder(response.Body).Decode(&apiError); err != nil {
+		return fmt.Errorf("%w: %d", errors.ErrHTTPStatusNotValid, response.StatusCode)
```
qdm12: attach the entire body string to the error, since it's not json

internal/provider/providers/hostinger/provider.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+type recordContent struct {
+	Content string `json:"content"`
+}
+
+type apiError struct {
+	Error         string `json:"error"`
+	CorrelationID string `json:"correlation_id"`
+}
```
qdm12: define these inline the functions they are needed, given they're each used only in one place. no need to have them at global scope

internal/provider/providers/hostinger/provider.go
```diff
@@ -0,0 +1,218 @@
...
+
+func (p *Provider) setHeaders(request *http.Request) {
+	headers.SetUserAgent(request)
+	headers.SetContentType(request, "application/json")
+	headers.SetAccept(request, "application/json")
+	headers.SetAuthBearer(request, p.token)
+}
+
+func (p *Provider) Update(ctx context.Context, client *http.Client, ip netip.Addr) (newIP netip.Addr, err error) {
```
qdm12, line 122: I'm a bit confused, should we set https://developers.hostinger.com/#tag/dns-zone/PUT/api/dns/v1/zones/{domain}.body.overwrite to true? As in, does sending the current http request just keeps on creating duplicate RR entries?

qdm12 (commented):
Thanks! 👍 Just a few minor comments, let me know if you want me to resolve them.

docs/hostinger.md
SpecialAro: Done ✔️ I updated `docs/hostinger.md` to follow the existing provider documentation format, including separate compulsory and optional parameter sections describing each setting.

internal/provider/providers/hostinger/provider.go
SpecialAro: Done ✔️ I moved the default TTL constant directly above the `extraSettings.TTL == 0` check.

internal/provider/providers/hostinger/provider.go
SpecialAro: Done ✔️ I replaced the single-case `switch` with a direct `if token == ""` check.

internal/provider/providers/hostinger/provider.go
SpecialAro: Done ✔️ I moved the update request types into `Update` and the API error response type into `handleAPIError`, since they are each only used within those functions.

internal/provider/providers/hostinger/provider.go
SpecialAro: Done ✔️ The response body is now read before unmarshalling, and the complete raw body is included in the returned error when it is not valid JSON.

internal/provider/providers/hostinger/provider.go
SpecialAro: Yes, I think `overwrite` should be set to `true` here. Hostinger documents that this deletes the existing resource records matching the supplied name and type before creating the new records. When it is `false`, existing TTLs are updated and the submitted records are appended.

Since this provider is updating the current A or AAAA value, using `true` prevents stale or duplicate address records from accumulating.

SpecialAro:
Thanks for the review!

I've addressed all the comments in the latest commit. I also verified that `overwrite: true` is intentional for this use case, since the updater needs to replace the existing matching A or AAAA records rather than append additional values.

Ready for another look, if you please! 😄

docs/hostinger.md
```diff
@@ -0,0 +1,38 @@
...
+## Configuration
+
+### Example
+
+```json
+{
+  "settings": [
+    {
+      "provider": "hostinger",
+      "domain": "example.com",
```
qdm12: just to fit with other documents

```suggestion
      "domain": "domain.com",
```

qdm12:
Thanks again!

PR #1148, opened 2026-06-03
fix(config): default binary program period option to 5m to match documentation

buddhaCode:
## Problem

The README documents the default `PERIOD` as `5m`, but the code uses `10m`. The documentation and the actual behavior disagree.

Code reference (current `master`):

```go
// internal/config/update.go
func (u *Update) setDefaults() {
	const defaultPeriod = 10 * time.Minute
	u.Period = gosettings.DefaultComparable(u.Period, defaultPeriod)
	const defaultCooldown = 5 * time.Minute
	u.Cooldown = gosettings.DefaultComparable(u.Cooldown, defaultCooldown)
}
```

I checked the history across releases `v2.4.0` through `v2.9.0` and `master`: the code default has always been `10m` (it was `params.Default("10m")` before the gosettings rework). So this looks like stale documentation rather than a code regression.

## Change

This PR only touches the README to align the docs with the code (`5m` -> `10m`) at the three relevant spots:

- The `PERIOD` row in the environment variables table.
- The Architecture section: "every period (10 minutes by default)".
- The Cloudflare special case: "every period (10 minutes by default), for each record".

The `UPDATE_COOLDOWN_PERIOD` default (`5m`) is left unchanged, since it correctly matches `defaultCooldown = 5 * time.Minute`.

No code is modified, so no Go tests are affected.

## Alternative

If `5m` was actually the intended default and the code is what's wrong, the fix would instead be to change `defaultPeriod` back to `5 * time.Minute` in `internal/config/update.go`. I went with the documentation fix as the lower-risk option, but happy to switch this to a code change instead if you prefer — your call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,7 +284,7 @@ Note that:
 | Environment variable | Default | Description |
 | --- | --- | --- |
 | `CONFIG` | | One line JSON object containing the entire config (takes precedence over config.json file) if specified |
-| `PERIOD` | `10m` | Default period of IP address check, following [this format](https://golang.org/pkg/time/#ParseDuration) |
+| `PERIOD` | `5m` | Default period of IP address check, following [this format](https://golang.org/pkg/time/#ParseDuration) |
 | `PUBLICIP_FETCHERS` | `all` | Comma separated fetcher types to obtain the public IP address from `http` and `dns` |
 | `PUBLICIP_HTTP_PROVIDERS` | `all` | Comma separated providers to obtain the public IP address (ipv4 or ipv6). See the [Public IP section](#public-ip) |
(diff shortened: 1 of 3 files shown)


qdm12 (changes requested):
Thanks! Actually it's the alternative, the intended is 5min. The Dockerfile has `PERIOD=5m` and ddns-updater is distributed as a single binary since "recently" as it was a container image only before this. Please revert and just change 10min -> 5min in the Go code, thanks!

buddhaCode:
Thanks for your review. I changed the code as requested and reverted the changes in the readme.

qdm12 (approved):

Issue #1039, opened 2025-10-07
Feature Request: Support for Hetzner DNS Migration to Hetzner Cloud Console


rueckert-digital:
### Description

On October 7, 2025, Hetzner announced that DNS Management is being migrated to the Hetzner Cloud Console:
🔗 [Hetzner News – DNS Beta](https://www.hetzner.com/news/dns-beta/)

This change introduces a new management interface and authentication model for DNS zones. It appears that existing API tokens and zone identifiers from the legacy DNS Console (https://dns.hetzner.com) are not compatible with the new Cloud Console (https://console.hetzner.com).

### Impact

1. API Tokens created in the old DNS Console are not migrated automatically.

  - [Old tokens](https://dns.hetzner.com/settings/api-token)
  - [New tokens](https://console.hetzner.com/projects/<project-id>/security/tokens)

2. Token format has changed — new tokens are significantly longer and likely tied to Hetzner Cloud project scope.

3. API endpoints have also changed:
 
  - [Legacy DNS API](https://dns.hetzner.com/api-docs/)
  - [New Cloud DNS API](https://docs.hetzner.cloud/reference/cloud#dns)

As a result, tools relying on the legacy Hetzner DNS API (such as ddns-updater) may no longer function properly once users migrate their domains to the Cloud Console.

### Request

Please add support for the new Hetzner Cloud DNS API, including:

- [ ] Updated authentication using new Cloud API tokens.
- [ ] Compatibility with zone identifiers under the Cloud Console structure.
- [ ] Optional backward compatibility with the legacy DNS API (if feasible).

### Related Issues / References

- #493
- #534

### Remarks

Fantastic work on ddns-updater — it’s an excellent tool and has worked flawlessly with the old Hetzner DNS API.
This feature would ensure continued compatibility with Hetzner’s new Cloud DNS environment going forward.

rueckert-digital:
Great work, @janolehuebner ...
Looking forward for merge and new docker image, @qdm12 😉

AltantisPegasus:
so i do unterstand it right, the issue is still open and we need the forked project?

Daxolion:
is the hetzner issue resolved?

@suntorytimed your image has a little problem, when using custom resolver adresses, like 1.1.1.1 for dns. 

when adding wildcard domains, the IP it stores is empty, causing rapid updates on the wildcard records.

update.json is filled, but the image tells itself, that the wildcard entry is empty and it has no IP, making it update again

[...]

PR #1149, opened 2026-06-03
fix(hetznercloud): poll zone action endpoint with adequate timeout

### LINKED ISSUE #1136: Bug: hetznercloud provider: action polling 404 marks successful updates as failed, container stays unhealthy
**TLDR**: hetznercloud provider treats successfully applied record updates as failed because action-polling endpoint returns 404, leaving the container permanently unhealthy after IP changes.

1. Is this urgent: No (functional workaround exists via `--no-healthcheck`)
2. DNS provider(s) you use: hetznercloud (Hetzner Cloud Console DNS, console.hetzner.com)
3. Program version: latest 2.10.0 (image pulled 2026-05-01)
4. What are you using to run the container: Unraid Docker (template-based, equivalent to docker run)
5. Extra information:

**What happens:**

1. Public IP changes (ISP reconnect)
2. ddns-updater detects the change and submits update to Hetzner
3. Hetzner accepts and applies the new IP - verified externally:
```
   $ dig @1.1.1.1 +short domain.tld.net
   188.174.63.100   # new IP, applied correctly
```
4. ddns-updater then polls the returned action ID for completion status
5. The action endpoint returns `404 not_found`
6. ddns-updater treats this as update failure
7. Healthcheck reports unhealthy, FailingStreak grows indefinitely (observed up to 22)
8. Reverse proxies (Traefik in my case) filter the unhealthy container out, breaking access to the WebUI
9. Manual docker restart resets the FailingStreak and brings the container back to healthy (since at startup, public IP and DNS already match), but the bug re-triggers on the next IP change. Only --no-healthcheck provides a stable workaround by removing the broken healthcheck from the equation entirely - the underlying action-polling bug remains.

**Suggested fix:**

For the `hetznercloud` provider, treat the update as synchronous (Hetzner's API applies the record write synchronously) and skip action polling entirely. Alternatively, treat 404 on action lookup as success since the record write itself returned a successful response.

**Distinct from:**

- Issue #1039 (hetznercloud provider config validation) - the provider does work for the actual record update; only action polling is broken
- Issue #851 (healthcheck timeout) - in my case the healthcheck endpoint responds correctly; the unhealthy status comes from ddns-updater's internal state, not from network-level timeout

Logs:
```log
2026-05-05T18:43:38+02:00 INFO ipv4 address of domain.tld is 188.174.60.180 and your ipv4 address is 188.174.63.100
2026-05-05T18:43:38+02:00 INFO ipv4 address of anotherdomain.tld is 188.174.60.180 and your ipv4 address is 188.174.63.100
2026-05-05T18:43:38+02:00 INFO Updating record [domain: domain.tld | owner: @ | provider: hetznercloud | ip: ipv4] to use 188.174.63.100
2026-05-05T18:43:38+02:00 ERROR updating record: handling action response: waiting for action to complete: HTTP status is not valid: 404: {{not_found action not found map[]}}
2026-05-05T18:43:38+02:00 INFO Updating record [domain: anotherdomain.tld | owner: @ | provider: hetznercloud | ip: ipv4] to use 188.174.63.100
2026-05-05T18:43:38+02:00 ERROR updating record: handling action response: waiting for action to complete: HTTP status is not valid: 404: {{not_found action not found map[]}}
```

Healthcheck output (from `docker inspect ddns-updater --format='{{json .State.Health}}'`):
```log
2026-05-05T19:00:54+02:00 ERROR program is unhealthy: record update failed: [domain: domain.tld | owner: @ | provider: hetznercloud | ip: ipv4]: failure (updating record: handling action response: waiting for action to complete: HTTP status is not valid: 404: {{not_found action not found map[]}})
Last success update: 2026-05-05 11:56:04 CEST
IP: 188.174.60.180
```

Same error on every retry. FailingStreak observed at 22 over ~22 minutes after a single IP change event. DNS records on Hetzner side were correctly updated to the new IP throughout this period.

Configuration file (**remove your credentials!**):
```json
{
  "settings": [
    {
      "provider": "hetznercloud",
      "zone_id": "REDACTED_NUMERIC_ID",
      "domain": "domain.tld",
      "owner": "@",
      "token": "REDACTED",
      "ttl": 60
    },
    {
      "provider": "hetznercloud",
      "zone_id": "REDACTED_NUMERIC_ID",
      "domain": "anotherdomain.tld",
      "owner": "@",
      "token": "REDACTED",
      "ttl": 60
    }
  ]
}
```

Host OS: Unraid 7.x (Slackware-based, Linux kernel 6.x)

buddhaCode:
## What

The `hetznercloud` provider marks successful record updates as failed,
leaving the container permanently unhealthy after an IP change. This fixes two
distinct bugs in the asynchronous action handling.

Fixes #1136

## Background: server actions vs zone actions

The Hetzner Cloud API returns an asynchronous *action* (a background job with
an id and a status) for many operations, and you poll that action until it
reports `success`. Crucially, actions are scoped to the **resource type** they
belong to, and each type has its own lookup path `/{resource}/actions/{id}`:

- `/v1/servers/actions/{id}` — actions for **Cloud Servers** (VPS instances),
  a completely separate Hetzner Cloud product.
- `/v1/zones/actions/{id}` — actions for **DNS zones**, which is what this
  provider works with.

A DNS record update produces a **zone** action, so it must be queried on the
zone path. Querying it on the server path asks the VPS/server subsystem for an
id it has never seen → `404 not_found`.

## Root cause

Record updates via Zone RRSet operations (e.g. `set_records`) return a zone
action that `waitAction` then polls until it reaches `success`. Two things
were wrong:

1. **Wrong endpoint.** It polled the server actions endpoint
   `https://api.hetzner.cloud/v1/servers/actions/{id}` instead of the zone
   actions endpoint `https://api.hetzner.cloud/v1/zones/actions/{id}`, so the
   lookup always returned `404 not_found`. The record itself was already
   updated server-side, but the 404 was treated as an update failure.
   ([waitaction.go#L20](https://github.com/qdm12/ddns-updater/blob/64996182d84adfeb7a873ea32f76b50918f90970/internal/provider/providers/hetznercloud/waitaction.go#L20))

2. **Too short a timeout.** Even against the correct endpoint, zone actions
   take ~20s to reach `success`, but `waitAction` only polled 3 times with a
   1s sleep (~3s total) and then gave up with "did not complete after 3 tries".
   Fixing only the URL would just swap the 404 for a timeout error — the
   container would stay unhealthy.

The action path scheme (`/{resource}/actions/{id}` with `resource = "zones"`)
matches Hetzner's official [hcloud-go](https://github.com/hetznercloud/hcloud-go)
library.

## Fix

- Poll the zone actions endpoint.
- Poll every 5s for up to 60s (12 tries) to accommodate the real action
  duration, while keeping the existing success / error / completion checks.

## Verified against the live Hetzner Cloud API

- Old endpoint `…/v1/servers/actions/{id}` → `404 not_found`; new endpoint
  `…/v1/zones/actions/{id}` → `200` with `"command": "set_rrset_records"` and
  `"resources": [{ "type": "zone" }]`.
- Measured action completion time: ~18–21s (hence the 60s budget).
- The real `ddns-updater` binary, before: `ERROR … 404: not_found`; after:
  the update completes with no error (verified by reading the record back).

## Alternatives considered

1. **Skip the action poll (optimistic update).** As suggested in the issue,
   treat the write as synchronous and consider the update successful as soon
   as `set_records`/`add_records` returns `201`, without waiting for the
   action to reach `success`. The record value is in fact applied server-side
   before the action completes (confirmed by reading it back while the action
   is still `running`). This could be offered as an opt-in provider setting
   (e.g. `"skip_action_check": true`) for users who prefer speed/robustness
   over confirmation. It is not the default here because it gives up genuine
   confirmation that the zone change was accepted.
2. **Treat a 404 on the action lookup as success.** More robust against
   further API quirks, but masks real failures.

This PR keeps the (now working) status check as the default. Happy to add the
opt-in `skip_action_check` option as well if you'd like it in the same PR.

## Open question on the timing values

I went with a fixed 5s interval / 60s total, which comfortably covers the
~20s I measured. Happy to change this if you'd prefer:

- a backoff strategy (e.g. 1s, 2s, 4s … capped) instead of a fixed interval,
- a longer overall budget, or
- making the interval/timeout configurable as optional fields in the provider
  settings (the provider already documents optional params like `ttl` in
  [docs/hetznercloud.md](https://github.com/qdm12/ddns-updater/blob/master/docs/hetznercloud.md)).

Let me know your preference and I'll adjust.

## Tests

`waitaction_test.go` covers the queried URL (the regression), the
`running → success` polling loop, the timeout path, and the success / 404 /
action-error cases. `go test ./...` and `golangci-lint run` pass.

### DIFF
--- a/internal/provider/providers/hetznercloud/update_wildcard_test.go
+++ b/internal/provider/providers/hetznercloud/update_wildcard_test.go
@@ -1,7 +1,6 @@
 package hetznercloud
 
 import (
-	"context"
 	"fmt"
 	"io"
 	"net/http"
@@ -39,7 +38,7 @@ func (rt *recordingTransport) RoundTrip(request *http.Request) (*http.Response, 
 		}, nil
 	}
 	return &http.Response{
-		StatusCode: response.statusCode,
+		StatusCode: int(response.statusCode),
 		Body:       io.NopCloser(strings.NewReader(response.body)),
 	}, nil
 }
@@ -60,7 +59,7 @@ func Test_Provider_Update_wildcard(t *testing.T) {
 		domain   = "example.com"
 		actionID = 12345
 	)
-	newIP := netip.MustParseAddr("185.28.78.5")
+	newIP := netip.AddrFrom4([4]byte{185, 28, 78, 5})
 
 	runningBody := fmt.Sprintf(`{"action":{"id":%d,"status":"running"}}`, actionID)
 	successBody := fmt.Sprintf(`{"action":{"id":%d,"status":"success"}}`, actionID)
@@ -126,7 +125,7 @@ func Test_Provider_Update_wildcard(t *testing.T) {
 				actionPollPeriod: time.Millisecond,
 			}
 
-			gotIP, err := provider.Update(context.Background(), client, newIP)
+			gotIP, err := provider.Update(t.Context(), client, newIP)
 			require.NoError(t, err)
 			assert.Equal(t, newIP, gotIP)
 
--- a/internal/provider/providers/hetznercloud/update_wildcard_test.go
+++ b/internal/provider/providers/hetznercloud/update_wildcard_test.go
@@ -1,0 +1,148 @@
+package hetznercloud
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/netip"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/qdm12/ddns-updater/pkg/publicip/ipversion"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// testToken is a dummy API token used across the hetznercloud provider tests.
+const testToken = "token"
+
+// recordingTransport records every request URL it sees and returns canned
+// responses keyed by "METHOD PATH". This lets us assert how the wildcard
+// owner "*" is turned into request URLs.
+type recordingTransport struct {
+	t         *testing.T
+	responses map[string]bodyResponse
+	gotURLs   []string
+}
+
+func (rt *recordingTransport) RoundTrip(request *http.Request) (*http.Response, error) {
+	rt.gotURLs = append(rt.gotURLs, request.URL.String())
+	key := request.Method + " " + request.URL.Path
+	response, ok := rt.responses[key]
+	if !ok {
+		rt.t.Errorf("unexpected request %s", key)
+		return &http.Response{
+			StatusCode: http.StatusNotImplemented,
+			Body:       io.NopCloser(strings.NewReader("{}")),
+		}, nil
+	}
+	return &http.Response{
+		StatusCode: response.statusCode,
+		Body:       io.NopCloser(strings.NewReader(response.body)),
+	}, nil
+}
+
+// Test_Provider_Update_wildcard is a regression test for the wildcard
+// rapid-update loop reported in
+// https://github.com/qdm12/ddns-updater/issues/1039 (comment by Daxolion).
+//
+// A wildcard owner "*" must be treated like any other owner: the "*" has to
+// appear literally in the rrset path (the Hetzner Cloud API returns 404 for
+// the percent-encoded "%2A"), and the asynchronous zone action returned by the
+// write must be polled on the zone actions endpoint. Getting either wrong made
+// the provider believe the update failed and retry it on every cycle.
+func Test_Provider_Update_wildcard(t *testing.T) {
+	t.Parallel()
+
+	const (
+		domain   = "example.com"
+		actionID = 12345
+	)
+	newIP := netip.MustParseAddr("185.28.78.5")
+
+	runningBody := fmt.Sprintf(`{"action":{"id":%d,"status":"running"}}`, actionID)
+	successBody := fmt.Sprintf(`{"action":{"id":%d,"status":"success"}}`, actionID)
+	createdActionBody := fmt.Sprintf(`{"action":{"id":%d,"status":"running"},`+
+		`"rrset":{"id":"*/A","name":"*","type":"A"}}`, actionID)
+
+	rrsetPath := "/v1/zones/" + domain + "/rrsets"
+	wildcardRRSetPath := rrsetPath + "/*/A"
+	zoneActionPath := fmt.Sprintf("/v1/zones/actions/%d", actionID)
+	serverActionPath := fmt.Sprintf("/v1/servers/actions/%d", actionID)
+
+	testCases := map[string]struct {
+		responses map[string]bodyResponse
+		// wantURLs are URLs that must have been requested.
+		wantURLs []string
+	}{
+		// The record does not exist yet: getRecord returns 404, so the provider
+		// creates the rrset and then polls the resulting zone action.
+		"create": {
+			responses: map[string]bodyResponse{
+				"GET " + wildcardRRSetPath: {http.StatusNotFound, `{"error":{"code":"not_found"}}`},
+				"POST " + rrsetPath:        {http.StatusCreated, createdActionBody},
+				"GET " + zoneActionPath:    {http.StatusOK, successBody},
+			},
+			wantURLs: []string{
+				"https://api.hetzner.cloud" + wildcardRRSetPath,
+				"https://api.hetzner.cloud" + rrsetPath,
+				"https://api.hetzner.cloud" + zoneActionPath,
+			},
+		},
+		// The record exists with a different IP: getRecord returns 200 and a
+		// stale value, so the provider replaces it via set_records and then
+		// polls the resulting zone action.
+		"set_records": {
+			responses: map[string]bodyResponse{
+				"GET " + wildcardRRSetPath: {
+					http.StatusOK,
+					`{"rrset":{"records":[{"value":"192.0.2.1"}]}}`,
+				},
+				"POST " + wildcardRRSetPath + "/actions/set_records": {http.StatusCreated, runningBody},
+				"GET " + zoneActionPath:                              {http.StatusOK, successBody},
+			},
+			wantURLs: []string{
+				"https://api.hetzner.cloud" + wildcardRRSetPath,
+				"https://api.hetzner.cloud" + wildcardRRSetPath + "/actions/set_records",
+				"https://api.hetzner.cloud" + zoneActionPath,
+			},
+		},
+	}
+
+	for name, testCase := range testCases {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			transport := &recordingTransport{t: t, responses: testCase.responses}
+			client := &http.Client{Transport: transport}
+
+			provider := &Provider{
+				domain:           domain,
+				owner:            "*",
+				ipVersion:        ipversion.IP4,
+				token:            testToken,
+				actionPollPeriod: time.Millisecond,
+			}
+
+			gotIP, err := provider.Update(context.Background(), client, newIP)
+			require.NoError(t, err)
+			assert.Equal(t, newIP, gotIP)
+
+			for _, wantURL := range testCase.wantURLs {
+				assert.Contains(t, transport.gotURLs, wantURL)
+			}
+
+			joinedURLs := strings.Join(transport.gotURLs, "\n")
+			// The wildcard must never be percent-encoded: the Hetzner Cloud API
+			// returns 404 for "%2A".
+			assert.NotContains(t, joinedURLs, "%2A")
+			assert.NotContains(t, joinedURLs, "%2a")
+			// The zone action must never be polled on the server actions
+			// endpoint (regression from issue #1136, which manifested as the
+			// wildcard rapid-update loop in issue #1039).
+			assert.NotContains(t, joinedURLs, serverActionPath)
+		})
+	}
+}
--- a/internal/provider/providers/hetznercloud/update_wildcard_test.go
+++ b/internal/provider/providers/hetznercloud/update_wildcard_test.go
@@ -1,0 +1,128 @@
+package hetznercloud
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"net/netip"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/qdm12/ddns-updater/pkg/publicip/ipversion"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Test_Provider_Update_wildcardOwner is a regression test for the wildcard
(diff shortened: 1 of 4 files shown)


buddhaCode:
Additional note: this also fixes the wildcard rapid-update loop reported in #1039 (comment by @Daxolion).

That report looked like a separate, wildcard-specific bug, but it is the same Zone-action polling issue fixed here. A wildcard owner `*` is handled like any other owner: `createRRSet` / `set_records` succeed server-side, but on `master` the resulting action was polled at `/v1/servers/actions/{id}` → `404 not_found` → the update was treated as failed and retried on every cycle.

I verified this against the live Hetzner Cloud API with a real wildcard owner, on both the create and the `set_records` paths: `master` loops, and with this PR the wildcard updates once and then converges. I also added a regression test (`update_wildcard_test.go`) that covers the wildcard owner (literal `*` in the rrset path, never `%2A`, and the zone actions endpoint).

I'm intentionally not adding a `Fixes #1039` keyword, since #1039 is the broader Hetzner Cloud migration tracking issue and should stay open. Relates to #1039.

buddhaCode:
Pushed two small follow-up commits while cross-checking the provider against the official API reference and `hcloud-go`:

- `test(hetznercloud)`: the `action_error` case in `waitaction_test.go` mocked the error as a bare string (`"error":"boom"`), but the API returns it as an object `{"code": ..., "message": ...}` ([reference](https://docs.hetzner.cloud/reference/cloud#tag/zone-actions/get_zone_action), and `schema.ActionError` in hcloud-go). The test now uses that realistic shape.
- `docs(hetznercloud)`: corrected the doc link anchor to the canonical `#tag/zone-actions/get_zone_action`.

While doing this I noticed a related, **pre-existing** point that is outside this PR's scope: `actionResponse.Action.Error` is typed as `any` (in `types.go`, unchanged by this PR), so on a real failure the `%v` formatting in `waitAction` would print something like `map[code:action_failed message:Action failed]` instead of a clean `action_failed: Action failed`. Typing it as a small struct (like the existing `errorResponse`) would make the error message readable.

Would you prefer I fold that fix into this PR, or keep it separate so this one stays focused on the zone-action endpoint? Happy either way.

internal/provider/providers/hetznercloud/waitaction_test.go
```diff
@@ -0,0 +1,165 @@
...
+// roundTripFunc lets us mock the HTTP client transport.
+type roundTripFunc func(request *http.Request) *http.Response
+
+func (f roundTripFunc) RoundTrip(request *http.Request) (*http.Response, error) {
+	return f(request), nil
+}
+
+// bodyResponse describes a single HTTP response returned by the mock transport.
+type bodyResponse struct {
+	statusCode int
```
qdm12: nit


```suggestion
type bodyResponse struct {
	statusCode uint
```

internal/provider/providers/hetznercloud/waitaction_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/qdm12/ddns-updater/internal/provider/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// roundTripFunc lets us mock the HTTP client transport.
```
qdm12:
```suggestion
```

internal/provider/providers/hetznercloud/waitaction_test.go
```diff
@@ -0,0 +1,165 @@
...
+	t.Parallel()
+
+	const actionID = 12345
+	runningBody := fmt.Sprintf(`{"action":{"id":%d,"status":"running"}}`, actionID)
+	successBody := fmt.Sprintf(`{"action":{"id":%d,"status":"success"}}`, actionID)
+
+	testCases := map[string]struct {
+		id           uint64
+		responses    []bodyResponse
+		wantQueries  int
```
qdm12: nit


```suggestion
		wantQueries  uint
```

internal/provider/providers/hetznercloud/waitaction_test.go
```diff
@@ -0,0 +1,165 @@
...
+			errWrapped:   errors.ErrUnsuccessful,
+			errSubstring: "action_failed",
+		},
+	}
+
+	for name, testCase := range testCases {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			var queries int
```
qdm12:
```suggestion
			var queries uint
```

internal/provider/providers/hetznercloud/waitaction_test.go
```diff
@@ -0,0 +1,165 @@
...
+					return &http.Response{
+						StatusCode: response.statusCode,
+						Body:       io.NopCloser(strings.NewReader(response.body)),
+					}
+				}),
+			}
+
+			// Use a tiny poll period to keep the test fast.
+			provider := &Provider{token: testToken, actionPollPeriod: time.Millisecond}
+			err := provider.waitAction(context.Background(), client, testCase.id)
```
qdm12:
```suggestion
			err := provider.waitAction(t.Context(), client, testCase.id)
```

internal/provider/providers/hetznercloud/waitaction_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+// Test_Provider_waitAction_timeout ensures waitAction gives up with an error
+// once the action did not complete within the allotted number of tries.
+func Test_Provider_waitAction_timeout(t *testing.T) {
+	t.Parallel()
+
+	const actionID = 12345
+	runningBody := fmt.Sprintf(`{"action":{"id":%d,"status":"running"}}`, actionID)
+
+	var queries int
```
qdm12:
```suggestion
	var queries uint
```

internal/provider/providers/hetznercloud/waitaction_test.go
```diff
@@ -0,0 +1,165 @@
...
+			queries++
+			return &http.Response{
+				StatusCode: http.StatusOK,
+				Body:       io.NopCloser(strings.NewReader(runningBody)),
+			}
+		}),
+	}
+
+	provider := &Provider{token: testToken, actionPollPeriod: time.Millisecond}
+	err := provider.waitAction(context.Background(), client, actionID)
```
qdm12:
```suggestion
	err := provider.waitAction(t.Context(), client, actionID)
```

internal/provider/providers/hetznercloud/update_wildcard_test.go
```diff
@@ -0,0 +1,148 @@
...
+// write must be polled on the zone actions endpoint. Getting either wrong made
+// the provider believe the update failed and retry it on every cycle.
+func Test_Provider_Update_wildcard(t *testing.T) {
+	t.Parallel()
+
+	const (
+		domain   = "example.com"
+		actionID = 12345
+	)
+	newIP := netip.MustParseAddr("185.28.78.5")
```
qdm12:
```suggestion
	newIP := netip.AddrFrom4([4]byte{185, 28, 78, 5})
```

qdm12 (approved):
Nice tests, good PR 👍 ! 20s to apply a dns record update is willddddd!! 🤷 

Just minor comment code suggestions I'll apply myself.

qdm12:
Actually I don't have push permissions on your repo, can you then apply those suggestions?

qdm12:
I would need you to push at least one commit as well, because CI somehow does not want to trigger on Github in this PR, probably a bug on Github's side.

buddhaCode:
All suggestions applied and pushed, which should also give CI something to trigger on.

On the 20s: it surprised me too. The ~18-21s I measured isn't a stable number. Some updates come back as success on the very first poll and are effectively synchronous, others take the full ~20s. My assumption is that it depends on how loaded Hetzner's zone backend is at that moment, but I can't verify that from the outside.

So the timeout has to cover the slow tail rather than the typical case, which is why I went with 60s. Fast updates return immediately anyway, so the larger budget only costs anything when Hetzner is actually being slow.

qdm12:
Can you update your branch with the master branch of this repo? There is a fix for the misspell action in the CI to run correctly. Also, from running locally there are 2 lint errors:

```
6.909 internal/provider/providers/hetznercloud/update_wildcard_test.go:41:18: G115: integer overflow conversion uint -> int (gosec)
6.909           StatusCode: int(response.statusCode),
6.909                          ^
6.909 internal/provider/providers/hetznercloud/waitaction_test.go:104:22: G115: integer overflow conversion uint -> int (gosec)
6.909                                           StatusCode: int(response.statusCode),
6.909                                                          ^
6.909 2 issues:
6.909 * gosec: 2
```

You can probably just add `//nolint:gosec` on both.

So let's make sure CI can run 😉 Thanks!!

qdm12:
With #1166 being merged, is this PR still needed? 🤔 Thanks!

buddhaCode:
Still needed, yes. #1166 fixed the endpoint. The timeout is the second bug and is still on master:

```go
const sleepDuration = time.Second
const tries = 3
```

That is a 3 second budget for a Zone action that takes around 20s to reach `success` (measured against the live API earlier in this PR). The poll now hits the right URL and gets a valid `200`, reads `"status":"running"` three times, then fails with:

```
unsuccessful result: action id 12345 did not complete after 3 tries
```

Same failed update as in #1136, different message. It is intermittent, because a write response that already reports `success` never reaches `waitAction`.

What is left here after #1166:

- 12 tries at 5s, so a 60s budget. Fast actions still return on the first poll.
- a test for the polling loop, since `Test_waitAction` serves one response per case and covers neither `running → success` nor the give-up path.
- a wildcard owner regression test for the update loop in #1039, over the create and the `set_records` branch.

Rebased onto master and force pushed, so CI has something to trigger on. The tests now reuse your `rewriteTransport` helper and I dropped my cases that duplicate `Test_waitAction`.

Both G115 errors are gone, but not via `//nolint:gosec`. I dropped the `uint → int` conversion instead, so the mock's `statusCode` is `int`, which is what `WriteHeader` takes. That reverts your `uint` nit on that one field. Reason:

| variant | v2.11.4 (pinned) | v2.12.2 |
| --- | --- | --- |
| `uint` + conversion | G115 ×2 | clean |
| `uint` + conversion + `//nolint:gosec` | clean | `nolintlint` ×2 |
| `int`, no conversion | clean | clean |

`//nolint:gosec` passes CI today, but breaks once the pinned version is bumped, since gosec stops reporting those lines and `nolintlint` then flags the directive as unused. Happy to switch back if you prefer. The poll counters stayed `uint`.

`go test -race ./...` and `golangci-lint run` are clean on both linter versions.

PR #1166, opened 2026-07-01
fix(hetznercloud): poll zone action endpoint instead of server actions

legend813:
## Problem

`waitAction` polls the action returned by `set_records` / `createRRSet` at:

```go
url := fmt.Sprintf("https://api.hetzner.cloud/v1/servers/actions/%d", id)
```

That action belongs to a **zone**, not a **server**. The `servers/actions` collection does not contain the id, so Hetzner responds with `404 not_found` ("action not found"). `handleErrorResponse` turns this into an error, so every update of an **existing** record is reported as `Failure` (and no last-IP is persisted), even though the `set_records`/create action itself succeeds server-side.

Records that never trigger a write (already up to date) are unaffected, which is why the bug only surfaces on the records that actually change IP.

## Fix

Poll the zone-scoped action endpoint instead:

```go
url := fmt.Sprintf("https://api.hetzner.cloud/v1/zones/actions/%d", id)
```

This matches the Hetzner Cloud API pattern `GET /{resource}/actions/{id}` used by the official `hetznercloud/hcloud-go` client (`ResourceActionClient.GetByID` → `"/" + resource + "/actions/%d"`, e.g. `/servers/actions/{id}` for servers, `/zones/actions/{id}` for zones).

## Test

Adds `waitaction_test.go` (httptest) asserting `waitAction` requests `/v1/zones/actions/{id}` and correctly surfaces an `error` action status.

`go test ./...` passes.

### DIFF
--- a/internal/provider/providers/hetznercloud/waitaction.go
+++ b/internal/provider/providers/hetznercloud/waitaction.go
@@ -18,7 +18,7 @@ func (p *Provider) waitAction(ctx context.Context, client *http.Client, id uint6
 	const sleepDuration = time.Second
 	const tries = 3
 	for range tries {
-		url := fmt.Sprintf("https://api.hetzner.cloud/v1/servers/actions/%d", id)
+		url := fmt.Sprintf("https://api.hetzner.cloud/v1/zones/actions/%d", id)
 		request, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


qdm12:
This may be a duplicate of #1149 - will review after 1149 gets merged.

buddhaCode:
Yes, this PR fixes the same issue as in 1149. But it doesn't touch the polling for the action result.

PR #1192, opened 2026-08-13
feat(provider): `rfc2136`

### LINKED ISSUE #325: Dynamic DNS Updates via TSIG (DynDNS)
Dyn provides an account level key which can be used to update DNS hosts instead of our HTTP-based DNS Update API. You can generate and obtain your key from your [TSIG account settings page](https://account.dyndns.com/profile/tsig.html). Your account must be a paid account to use this mechanism.

Please note: IPv6 is not supported via TSIG

TSIG updates are a mechanism to transport zone updates over a secured mechanism. This feature is available for paid accounts (DynDNS Pro and Dyn Standard DNS) and can be used with nsupdate or with dhcpd. For more information on this mechanism, please see [RFC 2845](http://tools.ietf.org/html/rfc2845) and the [Wikipedia page for TSIG](http://en.wikipedia.org/wiki/TSIG).

TTL values are only respected for Dyn Standard DNS hosts.

Using nsupdate for dynamic DNS TSIG Updates
To update a DNS server dynamically using TSIG for authorization, run nsupdate by doing the following:

$ nsupdate -d
> server update.dyndns.com
> zone $ZONE
> key $KEY_NAME $KEY_HMAC
> update add $HOST.$ZONE 60 A 10.0.0.1
> send
> quit
For Dynamic DNS hosts, $ZONE should match the host name (3rd level domain) to be updated (eg. myhost.dyndns.org)

### LINKED ISSUE #781: Add support for RFC 2136
I can not find that you offer support for RFC 2136? Dynamic dns updates implemented by Bind, Powerdns and others.

https://datatracker.ietf.org/doc/html/rfc2136
https://doc.powerdns.com/authoritative/dnsupdate.html

### LINKED ISSUE #1155: Feature request: RFC2136 provider
1. What's the feature?
A DNS provider according to RFC2136 dynamic DNS updates ( https://datatracker.ietf.org/doc/html/rfc2136 )
2. Extra information
RFC 2136 dynamic DNS updates allow for a standardized way to provide DNS updates, which keeping all infra local. It would greatly appreciated if you would consider such a provider!

francisrath:
Adds a provider updating records over DNS UPDATE messages ([RFC 2136](https://datatracker.ietf.org/doc/html/rfc2136)), optionally signed with a TSIG key. It works against any authoritative name server accepting dynamic updates, such as BIND, Knot DNS or PowerDNS.

Closes #781
Closes #1155
Closes #325

#325 asks for Dyn's TSIG updates, which is RFC 2136 with the server pinned to `update.dyndns.com`, so a configurable server covers it too.

## One thing worth flagging

This is the first provider that does not use HTTP, so the `*http.Client` passed to `Update` is unused. No interface change was needed, but it is a precedent, so say the word if you would rather it were done differently. `github.com/miekg/dns` was already a direct dependency, so `go.mod` is untouched.

## Design notes

- `zone` is optional and defaults to `domain`, which is right whenever the domain is the zone apex. It can be set for a domain sitting in a differently named zone.
- TSIG is optional. Without a key name the update is sent unsigned, for servers authorizing by address ACL instead.
- The update deletes the record set and adds the new record in a single message, so the server applies it atomically and creates the record if it does not exist yet.
- Updates go over TCP, since they are infrequent and it avoids truncated responses for signed messages.
- Rcodes map onto the existing errors: NOTAUTH to `ErrAuth`, REFUSED to `ErrBadRequest`, NOTZONE/NXDOMAIN to `ErrZoneNotFound`, SERVFAIL to `ErrDNSServerSide`.

## Testing

Build the image with:

```sh
docker build -t qmcgaw/ddns-updater:rfc2136 https://github.com/francisrath/ddns-updater.git#rfc2136
```

And then run with image `qmcgaw/ddns-updater:rfc2136` and a configuration such as:

```json
{
  "settings": [
    {
      "provider": "rfc2136",
      "domain": "example.com",
      "owner": "home",
      "server": "ns1.example.com:53",
      "tsig_key_name": "ddns-key",
      "tsig_secret": "base64secret==",
      "ip_version": "ipv4"
    }
  ]
}
```

[Configuration documentation](https://github.com/francisrath/ddns-updater/blob/rfc2136/docs/rfc2136.md)

Verified against BIND 9.20 in Docker: updating an existing record, creating a missing one, AAAA records, the zone apex, unsigned updates on a permissive zone, and the rejection paths (wrong TSIG secret, unsigned update on a key protected zone, unknown zone). Also ran the updater itself against it end to end.

If it is useful, I kept those tests on a branch stacked on this one, behind an `integration` build tag with a BIND container definition in `testdata`: https://github.com/francisrath/ddns-updater/tree/rfc2136-integration. I left them out of this PR since nothing else in the repo carries test fixtures, but happy to fold them in if you want them.

### DIFF
--- a/docs/rfc2136.md
+++ b/docs/rfc2136.md
@@ -12,7 +12,6 @@ Updates records on any authoritative name server accepting [RFC 2136](https://da
     {
       "provider": "rfc2136",
       "domain": "domain.com",
-      "owner": "@",
       "server": "ns1.domain.com:53",
       "tsig_key_name": "ddns-key",
       "tsig_secret": "base64encodedsecret==",
@@ -28,7 +27,7 @@ Updates records on any authoritative name server accepting [RFC 2136](https://da
 ### Compulsory parameters
 
 - `"domain"` is the domain to update. It can be `example.com` (root domain) or `sub.example.com` (subdomain of `example.com`).
-- `"server"` is the address of the name server accepting the dynamic updates. The port defaults to `53` if it is not specified. IPv6 addresses must be enclosed in square brackets, for example `[2001:db8::1]:53`.
+- `"server"` is the address of the name server accepting the dynamic updates, port included, for example `ns1.domain.com:53`. IPv6 addresses must be enclosed in square brackets, for example `[2001:db8::1]:53`.
 
 ### Optional parameters
 
--- a/docs/rfc2136.md
+++ b/docs/rfc2136.md
@@ -1,0 +1,70 @@
+# RFC 2136
+
+Updates records on any authoritative name server accepting [RFC 2136](https://datatracker.ietf.org/doc/html/rfc2136) dynamic DNS updates, such as [BIND](https://www.isc.org/bind/), [Knot DNS](https://www.knot-dns.cz/) or [PowerDNS](https://doc.powerdns.com/authoritative/dnsupdate.html). Updates are optionally authenticated with a [TSIG](https://datatracker.ietf.org/doc/html/rfc8945) key.
+
+## Configuration
+
+### Example
+
+```json
+{
+  "settings": [
+    {
+      "provider": "rfc2136",
+      "domain": "domain.com",
+      "owner": "@",
+      "server": "ns1.domain.com:53",
+      "tsig_key_name": "ddns-key",
+      "tsig_secret": "base64encodedsecret==",
+      "tsig_algorithm": "hmac-sha256",
+      "ttl": 300,
+      "ip_version": "ipv4",
+      "ipv6_suffix": ""
+    }
+  ]
+}
+```
+
+### Compulsory parameters
+
+- `"domain"` is the domain to update. It can be `example.com` (root domain) or `sub.example.com` (subdomain of `example.com`).
+- `"server"` is the address of the name server accepting the dynamic updates. The port defaults to `53` if it is not specified. IPv6 addresses must be enclosed in square brackets, for example `[2001:db8::1]:53`.
+
+### Optional parameters
+
+- `"tsig_key_name"` is the name of the TSIG key to sign the update with, as configured on the name server. If it is left unset, the update is sent unsigned, which only works if the name server authorizes the updater by other means, such as an address based ACL.
+- `"tsig_secret"` is the base64 encoded secret of the TSIG key. It is compulsory if `"tsig_key_name"` is set.
+- `"tsig_algorithm"` is the TSIG algorithm to use, and must match the algorithm configured on the name server. It can be `hmac-sha1`, `hmac-sha224`, `hmac-sha256`, `hmac-sha384` or `hmac-sha512`. It defaults to `hmac-sha256`.
+- `"zone"` is the zone to send the update to. It defaults to the `"domain"` value, which is correct in most cases. Set it if the domain to update sits in a zone with a different name, for example when updating `host.sub.example.com` in the `example.com` zone.
+- `"ttl"` is the TTL in seconds of the record to update. It defaults to `300`.
+- `"ip_version"` can be `ipv4` (A records), or `ipv6` (AAAA records) or `ipv4 or ipv6` (update one of the two, depending on the public ip found). It defaults to `ipv4 or ipv6`.
+- `"ipv6_suffix"` is the IPv6 interface identifier suffix to use. It can be for example `0:0:0:0:72ad:8fbb:a54e:bedd/64`. If left empty, it defaults to no suffix and the raw temporary IPv6 address of the machine is used in the record updating. You might want to set this to use your permanent IPv6 address instead of your temporary IPv6 address.
+
+## Domain setup
+
+The steps below are for BIND, but the same concepts apply to other name servers.
+
+1. Generate a TSIG key, for example with `tsig-keygen -a hmac-sha256 ddns-key`, which outputs a block such as:
+
+    ```conf
+    key "ddns-key" {
+        algorithm hmac-sha256;
+        secret "base64encodedsecret==";
+    };
+    ```
+
+1. Add that key block to your BIND configuration.
+1. Allow the key to update your zone:
+
+    ```conf
+    zone "domain.com" {
+        type primary;
+        file "/var/lib/bind/db.domain.com";
+        allow-update { key "ddns-key"; };
+    };
+    ```
+
+1. Reload the configuration with `rndc reload`.
+1. Set `"tsig_key_name"` and `"tsig_secret"` in your configuration to the key name and secret generated above.
+
+Note the updater replaces the entire A or AAAA record set of the domain name with a single record, and creates the record if it does not exist yet. Updates are sent over TCP.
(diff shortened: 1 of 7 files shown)


docs/rfc2136.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Example
+
+```json
+{
+  "settings": [
+    {
+      "provider": "rfc2136",
+      "domain": "domain.com",
+      "owner": "@",
```
qdm12: the code now understands the owner/domain from just the domain string

```suggestion
      "domain": "domain.com",
```

docs/rfc2136.md
```diff
@@ -0,0 +1,70 @@
...
+      "ipv6_suffix": ""
+    }
+  ]
+}
+```
+
+### Compulsory parameters
+
+- `"domain"` is the domain to update. It can be `example.com` (root domain) or `sub.example.com` (subdomain of `example.com`).
+- `"server"` is the address of the name server accepting the dynamic updates. The port defaults to `53` if it is not specified. IPv6 addresses must be enclosed in square brackets, for example `[2001:db8::1]:53`.
```
qdm12: to simplify things let's just enforce specifying a port always

internal/provider/providers/rfc2136/provider.go
```diff
@@ -0,0 +1,242 @@
...
+		tsigKeyName:   providerSpecificSettings.TSIGKeyName,
+		tsigSecret:    providerSpecificSettings.TSIGSecret,
+		tsigAlgorithm: providerSpecificSettings.TSIGAlgorithm,
+		ttl:           providerSpecificSettings.TTL,
+	}, nil
+}
+
+// addressWithDefaultPort adds the default DNS port to the server address
+// given, if the address does not already specify a port.
+func addressWithDefaultPort(server string) (address string) {
```
qdm12: let's enforce to specify the port, to simplify code/documentation/usage to a single way to do it

internal/provider/providers/rfc2136/provider.go
```diff
@@ -0,0 +1,242 @@
...
+// Update replaces the A or AAAA record set of the domain name with a single
+// record pointing at the IP address given, using a DNS update message as
+// specified in https://datatracker.ietf.org/doc/html/rfc2136.
+// The record is created if it does not exist yet.
+// The HTTP client is unused since the update is sent over DNS.
+func (p *Provider) Update(ctx context.Context, _ *http.Client, ip netip.Addr) (newIP netip.Addr, err error) {
+	message := p.newUpdateMessage(ip)
+
+	// TCP is used instead of UDP to avoid truncated responses, since
+	// updates are infrequent enough for the extra round trips not to matter.
```
qdm12, line 175: 👍 thanks for the explanation!

internal/provider/providers/rfc2136/provider.go
```diff
@@ -0,0 +1,242 @@
...
+		Class:  dns.ClassINET,
+		Ttl:    p.ttl,
+	}
+
+	var record dns.RR
+	if ip.Is6() {
+		header.Rrtype = dns.TypeAAAA
+		record = &dns.AAAA{Hdr: header, AAAA: net.IP(ip.AsSlice())}
+	} else {
+		record = &dns.A{Hdr: header, A: net.IP(ip.AsSlice())}
```
qdm12, line 212: nit just for simplicity/easier reading


```suggestion
	header := dns.RR_Header{
		Name:   dns.Fqdn(utils.BuildDomainName(p.owner, p.domain)),
		Class:  dns.ClassINET,
		Ttl:    p.ttl,
	}

	var record dns.RR
	if ip.Is6() {
		header.Rrtype = dns.TypeAAAA
		record = &dns.AAAA{Hdr: header, AAAA: net.IP(ip.AsSlice())}
	} else {
		header.Rrtype = dns.TypeA
		record = &dns.A{Hdr: header, A: net.IP(ip.AsSlice())}
```

I was thinking initially uh oh this does not account for AAAA? 🤔 😄

internal/provider/providers/rfc2136/provider.go
```diff
@@ -0,0 +1,242 @@
...
+	case dns.RcodeRefused:
+		return errors.ErrBadRequest
+	case dns.RcodeNotZone, dns.RcodeNameError:
+		return errors.ErrZoneNotFound
+	case dns.RcodeFormatError:
+		return errors.ErrBadRequest
+	case dns.RcodeServerFailure:
+		return errors.ErrDNSServerSide
+	default:
+		return errors.ErrUnknownResponse
```
qdm12:
```suggestion
		return fmt.Errorf("%w", errors.ErrUnknownResponse)
```

internal/provider/providers/rfc2136/provider.go
```diff
@@ -0,0 +1,242 @@
...
+	case dns.RcodeNotAuth:
+		return errors.ErrAuth
+	case dns.RcodeRefused:
+		return errors.ErrBadRequest
+	case dns.RcodeNotZone, dns.RcodeNameError:
+		return errors.ErrZoneNotFound
+	case dns.RcodeFormatError:
+		return errors.ErrBadRequest
+	case dns.RcodeServerFailure:
+		return errors.ErrDNSServerSide
```
qdm12: nit use `fmt.Errorf` to enforce callers to use `errors.Is` instead of `==`


```suggestion
		return fmt.Errorf("%w", errors.ErrAuth)
	case dns.RcodeRefused:
		return fmt.Errorf("%w", errors.ErrBadRequest)
	case dns.RcodeNotZone, dns.RcodeNameError:
		return fmt.Errorf("%w", errors.ErrZoneNotFound)
	case dns.RcodeFormatError:
		return fmt.Errorf("%w", errors.ErrBadRequest)
	case dns.RcodeServerFailure:
		return fmt.Errorf("%w", errors.ErrDNSServerSide)
```

internal/provider/providers/rfc2136/provider_test.go
```diff
@@ -0,0 +1,171 @@
...
+				require.ErrorIs(t, err, testCase.expectedErr)
+				return
+			}
+
+			require.NoError(t, err)
+			assert.Equal(t, testCase.expectedZone, provider.zone)
+			assert.Equal(t, testCase.expectedTTL, provider.ttl)
+			assert.Equal(t, testCase.expectedSrv, provider.server)
+			assert.Equal(t, testCase.expectedAlgo, provider.tsigAlgorithm)
+			assert.Equal(t, testCase.expectedKeyNm, provider.tsigKeyName)
```
qdm12: nit assert the provider against an expectedProvider struct directly, no need to assert field by field

qdm12 (approved):
Nice PR 💪 🎖️ Just some minor comments 😉 thanks!!!

qdm12:
If you can also please merge in the master branch in your branch, there is a fix for the CI 🤷 thanks!

francisrath:
Thanks for your quick and thorough review @qdm12 - I rebased the end-to-end test locally and reran it also. I assume you don't want to have that in the repo?

